Issue #57, opened 2023-03-23
Confirmation dialog for "Delete All" history button


sintezcs:
Please add a confirmation dialog for the "Delete All history" button. I accidentally clicked it, and it removed all my history, which I didn't want to do. 
A confirmation for deleting a single conversation may also be useful.

carlrobertoh:
Hi @sintezcs, I'm sorry to hear that.
This was actually something that I was expecting to happen to someone sooner or later.
I will try to add the dialog in the upcoming days.

FYI: This is a good time to update the plugin to 1.7.0 - https://plugins.jetbrains.com/plugin/21056-codegpt/versions/stable/310112

PR #59, opened 2023-03-23
Confirmation dialogs for deleting conversations

sintezcs:
A fix for this issue: https://github.com/carlrobertoh/CodeGPT/issues/57

Show a confirmation dialog when clicking the "Delete Conversation" and "Delete All" buttons in the "Conversation History" tab. This will prevent the accidental removal of conversations that may be important for a user.


carlrobertoh:
Thank you!

PR #64, opened 2023-03-28
Change how placeholder text is displayed and hidden

arcticlinux:
- based on the IntelliJ Platform UI Guidelines https://jetbrains.design/intellij/controls/input_field/#placeholder "Hide the placeholder when the user starts typing, not when the input field gets the focus."
- remove focus event and listener

I was running into an issue where if you paste by using middle click in Linux, the placeholder text would not be erased first, and the pasted text and the placeholder end up being combined.

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/toolwindow/components/TextArea.java
+++ b/src/main/java/ee/carlrobert/codegpt/toolwindow/components/TextArea.java
@@ -8,18 +8,16 @@ import com.intellij.ui.components.JBTextArea;
 import com.intellij.util.ui.JBUI;
 import icons.Icons;
 import java.awt.event.ActionListener;
-import java.awt.event.FocusEvent;
-import java.awt.event.FocusListener;
 import javax.swing.JButton;
 import javax.swing.JScrollPane;
 
 public class TextArea extends JBTextArea {
 
   public TextArea(Runnable onSubmit, JScrollPane textAreaScrollPane) {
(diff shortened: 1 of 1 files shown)


arcticlinux:
I don't think the `StatusVisibleFunction` part of this pull request is required any longer, I will test and update the branch.

arcticlinux:
I have reduced the pull request and commit to the minimum change required, it should be simpler to review now.

carlrobertoh:
Foreground color needs to be set to `JBColor.BLACK`, otherwise the inserted text color remains gray.
Previously it was handled inside the focus callbacks.
Screen Shot 2023-03-28 at 23 21 56

arcticlinux:
The placeholder text is styled gray by default so the setting of the color isn't required, and the foreground color will be the theme default, which should be black or white.  I don't think it needs to be set explictly, but it could be done that way with something like:

```
setForeground(JBColor.BLACK);
SimpleTextAttributes JB_GRAY = new SimpleTextAttributes(STYLE_PLAIN, JBColor.GRAY);
getEmptyText().setText("Ask me anything...", JB_GRAY);
```

carlrobertoh:
I think that's fine for now.
I'm working on a bigger change to put the response generation inside the embedded browser to have a proper code blocks and to be visually more pleasing and later on I might replace the TextArea as well.
Screen Shot 2023-03-29 at 11 48 53

Issue #72, opened 2023-04-02
Add a custom prompt on the left side of the input box.


MiniFullC:
Very useful! Thank you to the author~ I have a suggestion: 
Add a custom prompt on the left side of the input box, as shown in the picture:

![image](https://user-images.githubusercontent.com/37063727/229363932-819a647c-107b-4ff0-81de-84861c051d6b.png)

You can look at [awesome-chatgpt-prompts](https://github.com/f/awesome-chatgpt-prompts) to find interesting features to import into the app.

It would be great to be able to import prompts from a custom CSV file, just like the **prompts.csv** file in this repository.

The suggestion is to use the forward slash shortcut key to conveniently use prompts.

carlrobertoh:
Hi, thank you for the feature request!

This seems like a nice idea altho some of these prompts are rather useless when it comes to developer experience (e.g. "act as a rival chess player"). Also there are some plans replacing the Java Swing components (JTextArea and JButton) with browser based components and that might complicate few things. 

However, I'm planning on adding a support for overriding the request params and these system lvl prompts can be one of the things that user could override altho you don't get that nice user experience as shown in the picture.

PR #83, opened 2023-04-11
Editable system role 

FxProjects:
image
image
image
I added the ability to change the parameter for the system role. which allows you to more accurately and conveniently configure the bot and specify its behavioral data

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/action/ActionsUtil.java
+++ b/src/main/java/ee/carlrobert/codegpt/action/ActionsUtil.java
@@ -12,6 +12,7 @@ import com.intellij.openapi.project.Project;
 import ee.carlrobert.codegpt.util.FileUtils;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 
 public class ActionsUtil {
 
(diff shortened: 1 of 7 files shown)


FxProjects:
image
image
image
And add abiluty to define by user custom hotkeys for actions 
image

carlrobertoh:
Thank you for contributing!

First of all, I'd love if these two distinct features have a separate pull requests.

Secondly, the system role messages aren't really part of the toolwindow's logic i.e. it should be configurable while overriding the API request params (which in fact doesn't exist today). Request param overriding should be done inside the plugin's settings, similarly how selected code actions are configured. Also, there is an [open ticket](https://github.com/carlrobertoh/CodeGPT/issues/72) for that.

Thirdly, the key mappings should also include actions such as submitting the input, copying/replacing the generated code, navigating between main tabs(Chat/Conversation History), selecting the previous conversation etc. However, I'm also happy to accept the partial logic for it.

Lastly, try to keep the code changes as minimal as possible (e.g. don't commit the file if it only includes the code formatting). Also, I'm trying to follow the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html) since the plugin doesn't specify any code style rules.

FxProjects:
To be honest, I don’t write in java at all, but I’ll definitely read and improve my skills for the future and try to comply with these standards.
I understand that here you can really make a huge functionality and it will take a lot of time and I have no right to demand these functions from you, so I tried to do it myself as it turned out and shared who might need it as developments.

carlrobertoh:
That's perfectly fine! 🙂 

For example, I'm happy to accept the logic for the actions key mappings if you tidy up the code a bit and lose the other functionality(or create another PR for it).

jmckenzie-dev:
`I'm happy to accept the logic for the actions key mappings`
Just want to chime in and say: this would make me a _very happy ideaVim user_. 😀 

Appreciate the plugin and appreciate the work on adding these key mappings. Super impressed with all of it!

carlrobertoh:
@jmckenzie-dev I will try to add the basic key mappings next week.

carlrobertoh:
@jmckenzie-dev Key mappings are now available in version 1.10.7 (pending Jetbrains approval).

jmckenzie-dev:
🎉 

That's freaking awesome. Thanks!

PR #111, opened 2023-05-18
Add custom API host to allow users to set up reverse proxy for OpenAI…

funv5:
To make it easier for users to set up reverse proxy for OpenAI's custom domain API host, we can add a custom API host text box on the settings page, allowing users to enter the API host they want to use.


carlrobertoh:
@ydf This functionality already exist in the latest version.

PR #212, opened 2023-09-21
anymous telemetry based on redhat

siilats:


nicdesousa:
Why do you need telemetry?
- Is what you're collecting here (or in the future) _actually_ going to benefit the users of this plugin? (Since bug reporting, feature requests, etc., is already available.)
- There's a slippery slope of having users opt-in to the _initial_ data collection set (by persisting telemetry.enabled = true), and then expanding the collection scope in the future _without_ obtaining consent for the additional data...
- Correct me if I'm wrong, but are you doing this to assess commercialization of the plugin?

siilats:
> Why do you need telemetry?
> 
> * Is what you're collecting here (or in the future) _actually_ going to benefit the users of this plugin? (Since bug reporting, feature requests, etc., is already available.)
> * There's a slippery slope of having users opt-in to the _initial_ data collection set (by persisting telemetry.enabled = true), and then expanding the collection scope in the future _without_ obtaining consent for the additional data...
> * Correct me if I'm wrong, but are you doing this to assess commercialization of the plugin?

In practice when the usage of the plugin becomes more popular and the use rbase becomes wider, the new users will stop reporting bugs because they just don’t do it. Currently many users are sophisticated and file issues, but we have certain bugs on certain Jetbrains products that you only see on telemetry. We use the standard telemetry library from Redhat from https://github.com/redhat-developer/intellij-redhat-telemetry and the plugin is open source. The telemetry implementation is in this separate pull request so you can easily see what it does. Basically it says request from pycharm to [you.com](http://you.com/) at 8pm was successful. It allows us to monitor if an LLM model provider goes down. Finally we are working on running our own instance of the Facebook code-llama2 for free, so we need telemetry for  preventing abuse there. I hope that’s ok, if you have experience with some Redhat settings please let us know, I don’t see this telemetry.enabled = true anywhere

nicdesousa:
I really like this plugin and was thinking of contributing..., so thank you for taking the time to provide an explanation/motivation.
- May I suggest that you update the PR description with this explanation?

>The telemetry implementation is in this separate pull request so you can easily see what it does.

Indeed, which I greatly appreciate. I've seen other plugins where telemetry collection wasn't implemented with the level of transparency that you're demonstrating.

>I hope that’s ok, if you have experience with some Redhat settings please let us know, 

It certainly makes a lot more sense now that I understand the context. 
I think this is OK, provided that telemetry collection remains an _explicit_ opt-in feature.

>I don’t see this telemetry.enabled = true anywhere

That's just pseudo-code referring to the persisted configuration for `ee.carlrobert.codegpt.telemetry.ui.preferences.TelemetryConfigurable`.

PR #223, opened 2023-10-02
Renamed `id` property on `Conversation` to `localId`

strangedreamsNY:
Changing the name here clarifies the code to indicate that a Conversation's 'id' is a local one and not one provided by the users selected AI service.


strangedreamsNY:
I tried to look for any mappings that would indicate this would cause issues against a logging back-end, but it looks like this property was being mapped to 'conversationId' when sent as a Telemetry event, so just updated the property retrieval in the tests.

carlrobertoh:
I'm not sure if I fully understand the problem or the necessity for this change.

The `id` is just an unique identifier for conversations stored *locally* on your file system. There is no relation with any third-party services, or even if there were a connection, a single id couldn't be referenced across multiple different systems.

Finally, this change is not backward compatible. Deserialization of existing conversations will fail because the field `id` no longer exists.

PR #228, opened 2023-10-09
Fix StreamParser with text-code-text message sequence

tautcony:
Current parsers can only support `code-text`, `text-code`, `text`, `code`.

The following message(`text-code-text`) will be treated as text and will cause the HeaderActions of the code fragment not to be displayed.

````
foo
```
ls -la
```
bar
````

Switching to split markdown blocks with `MarkdownUtils.splitCodeBlocks` that are consistent with usage in the `ChatMessageResponseBody` would fix this kind of problem and ensure consistency throughout the project.


carlrobertoh:
`StreamParser` is used throughout the completion request and it receives partial messages as input.
I'm not sure if you've tested the changes, but they don't work because it always evaluates whether the partial message is code or text.

The current implementation is messy, I know. 
I would greatly appreciate it if someone could rewrite the logic. 🙂

tautcony:
Understood, this may only occur with non-stream API calls, this pr would be better to skip for now.

PR #229, opened 2023-10-10
Fix memory leak in extensions

tautcony:
Currently the following error occurs when compiling the plugin.

```
Memory leak detected: 'ee.carlrobert.codegpt.settings.SettingsConfigurable@4bcf0b26' (class ee.carlrobert.codegpt.settings.SettingsConfigurable) was registered in Disposer as a child of 'ROOT_DISPOSABLE' (class com.intellij.openapi.util.Disposer$2) but wasn't disposed.
Register it with a proper parentDisposable or ensure that it's always disposed by direct Disposer.dispose call.
See https://jetbrains.org/intellij/sdk/docs/basics/disposers.html for more details.
```

According to the [documentation](https://plugins.jetbrains.com/docs/intellij/disposers.html#automatically-disposed-objects), extensions registered in `plugin.xml` are not automatically disposed causing this error.

So defining [a disposable parent](https://plugins.jetbrains.com/docs/intellij/disposers.html#choosing-a-disposable-parent) and disposing it in `disposeUIResources` should fix this error.

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/settings/ServiceSelectionForm.java
+++ b/src/main/java/ee/carlrobert/codegpt/settings/ServiceSelectionForm.java
@@ -1,8 +1,8 @@
 package ee.carlrobert.codegpt.settings;
 
+import com.intellij.openapi.Disposable;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.ui.ComboBox;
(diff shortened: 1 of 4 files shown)


carlrobertoh:
Thank you for contributing, I will review the PRs as soon as possible.

PR #295, opened 2023-11-28
Local LLM: Added empty check for `Additional parameters` field

ViktorGozhiy:
After adding the **Additional parameters** field, a problem arose with running local models. The problem is that if the **Additional parameters** field is left empty, then an empty parameter is added to the command line. It looks like this:
```./server -m /home/user/path/to/model/codellama-13b-instruct.Q4_K_M.gguf -c 2048 --port 39193 -t 20 ""```
Pay attention to the end of the line: `""`
Thus, the server does not start and on the screen we see the error: **Server Terminated**
This pull request fixes this issue.


carlrobertoh (approved):
Thank you!

Issue #303, opened 2023-12-06
Copilot style autocompletion


jihiggins:
### Describe the need of your request

Hi! Thanks for making this project, it's pretty neat so far.

I was wondering if it's on the roadmap to add copilot style autocomplete suggestions (or if there's a roadmap in general?) So far, nothing else I've looked at seems to do the copilot style thing while still allowing people to use their own api keys.

 I might also be able to help with this at some point, if contributors are needed

PR #308, opened 2023-12-15
Inline Autocompletion (WIP)

jihiggins:
Will add automatically generated inline autocomplete suggestions (#303)

Putting up an early draft for any discussion etc

Rough outline:
- [x] Inline suggestion inlays 
- [x] Action / tab complete
- [ ] Settings
  - [ ] Rules on when to generate a suggestion (end of line or after certain chars?)
  - [ ] End tokens (e.g. newline for only generating one line at a time)
  - [ ] Idle time before querying GPT
  - [ ] How much context around the cursor to send
  - [ ] Other forms of context?
  - [ ] Different setting for which model to use? (for things like using gpt4 for chat, gpt3.5 turbo for code completion to lower costs)
  - [ ] The prompt to use for generating the autocompletion text
- [ ] Query GPT completions for the autocompletion

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/actions/editor/InsertInlineTextAction.java
+++ b/src/main/java/ee/carlrobert/codegpt/actions/editor/InsertInlineTextAction.java
@@ -1,0 +1,41 @@
+package ee.carlrobert.codegpt.actions.editor;
+
+import com.intellij.openapi.actionSystem.DataContext;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.command.WriteCommandAction;
+import com.intellij.openapi.editor.Caret;
+import com.intellij.openapi.editor.Editor;
+import com.intellij.openapi.editor.Inlay;
+import com.intellij.openapi.editor.actionSystem.EditorAction;
+import com.intellij.openapi.editor.actionSystem.EditorActionHandler;
+import com.intellij.openapi.util.Key;
+import ee.carlrobert.codegpt.codecompletion.InlineCodeCompletionRenderer;
+import org.jetbrains.annotations.NotNull;
+
+public class InsertInlineTextAction extends EditorAction {
+    public InsertInlineTextAction(
+            Key<Inlay<InlineCodeCompletionRenderer>> inlayKey,
+            String text
+    ) {
+        super(new EditorActionHandler() {
+            @Override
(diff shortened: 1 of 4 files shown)


jihiggins:
It looks like one problem here is it's actually non-trivial to come up with a prompt that gives you the right sort of reply for this. (Or I'm just bad at coming up with the prompt, who knows 🥲) The closest I've gotten is by having it complete things normally and then doing a diff of some kind.

I think it might be worth figuring the prompt out before continuing to get into this, since the complexity of the prompt could impact what needs to get put together. (Especially since we can come up with prompts separately outside of the IDE with mock code, etc.)

carlrobertoh:
> It looks like one problem here is it's actually non-trivial to come up with a prompt that gives you the right sort of reply for this. (Or I'm just bad at coming up with the prompt, who knows 🥲) The closest I've gotten is by having it complete things normally and then doing a diff of some kind.
> 
> I think it might be worth figuring the prompt out before continuing to get into this, since the complexity of the prompt could impact what needs to get put together. (Especially since we can come up with prompts separately outside of the IDE with mock code, etc.)

We need to use a special type of prompt - Fill-in-the-middle (FIM).
Here's a bit of info how to to use it: https://ollama.ai/blog/how-to-prompt-code-llama
Available tokens: https://github.com/openai/tiktoken/blob/main/tiktoken_ext/openai_public.py#L3C17-L3C17
Example request:
```json
{
  "model": "gpt-4",
  "temperature": 0.0,
  "messages": [
    {
      "role": "user",
      "content": "<|fim_prefix|> def compute_gcd(x, y): <|fim_suffix|>return result <|fim_middle|>"
    }
  ]
}
```

jihiggins:
That looks like it'd work for Llama! OpenAI has a similar thing with `suffix` in the legacy completions API, but that's getting deprecated, and afaict the chat endpoints don't have anything similar?

carlrobertoh:
The same technique can be used for OpenAI Chat Completion API as well. Updated the example request.

jihiggins:
Ah, awesome, didn't realize it would just work that way. Seems to work with everything except for the new gpt-4-turbo model. Thanks!

PhilKes:
Hi, I took the liberty to try to implement a first draft for using FIM to get code-completion suggestions.
Since I cant push to your PR branch, I created a new PR: https://github.com/carlrobertoh/CodeGPT/pull/333

PR #336, opened 2024-01-20
fix: switch enabled state of custom model settings

HarukaMa:
The model path and prompt template settings were not being enabled / disabled when user starts or stops the server.


reneleonhardt:
@HarukaMa Can you please rebase your PR locally?
The file has since been moved 2 directories down from / to:
src/main/java/ee/carlrobert/codegpt/settings/service/LlamaModelPreferencesForm.java
src/main/java/ee/carlrobert/codegpt/settings/service/llama/form/LlamaModelPreferencesForm.java

HarukaMa:
I probably will need to check if other newly added local model settings need to be updated as well.

carlrobertoh:
It should be fixed. Please reopen if it isn't.

PR #337, opened 2024-01-20
fix: show custom model filename in chat panel

HarukaMa:
The model selection combo box above the prompt text area was always showing the hugging face model, even when the active selection is a custom model.


reneleonhardt:
@HarukaMa Can you check if the changes https://github.com/carlrobertoh/CodeGPT/commit/7c067d9edd546375be419907b068d1b003e0db25 since 2.3.0 fixed your problem?

HarukaMa:
Oh, yes, it has already been fixed some time ago but I forget to follow up here.

reneleonhardt:
No problem, glad that this bug has been fixed, thank you for reporting and providing an implementation 👍

PR #339, opened 2024-01-23
Switching models in the Model tab

vauns:
In the Model tab, when switching models and creating a new chat box, the icon and model inside the current chat box should not be modified


reneleonhardt:
@carlrobertoh This looks like a bug, the removed lines are still there. Can you double-check if this fix is still needed?

carlrobertoh:
Hmm, which removed lines are you referring to? I don't think this fix is needed anymore.

reneleonhardt:
The diff removed lines 151 and 152, I was wondering if is this bug is fixed with those still present:
https://github.com/carlrobertoh/CodeGPT/pull/339/files#diff-662da4888f3e65598f529191ee635da51faf0f22036f5791d0c1a84cd1647c79L151

That was basically the whole fix 😅
`Only set icon and text if there is no current conversation`

vauns:
When i have time i will show the problem with gif.

vauns:
https://github.com/carlrobertoh/CodeGPT/assets/14341441/a692dcc2-4107-4780-879d-8bccac2f3f76

When there is a current conversation, selecting another model will open a new chat, but it should not change the icon and model name of the previous chat. At the same time, the actual model of the previous conversation is correct (because it has not changed), but the icon and model name are incorrect and do not match, which may cause misleading information

vauns:
image

Considering the impact and correctness of historical sessions, the final code may look like this
`ModelComboBoxAction.java`
```
private void handleProviderChange(
        ServiceType serviceType,
        String label,
        Icon icon,
        Presentation comboBoxPresentation) {
    settings.setSelectedService(serviceType);

    var currentConversation = ConversationsState.getCurrentConversation();
    if (Objects.isNull(currentConversation) || currentConversation.getMessages().isEmpty()) {
        comboBoxPresentation.setIcon(icon);
        comboBoxPresentation.setText(label);
        if (Objects.isNull(currentConversation)) {
            ConversationService.getInstance().startConversation();
        } else {
            ConversationService.getInstance().updateConversation(currentConversation);
        }
    } else {
        onAddNewTab.run();
    }
}
```
`ConversationService.java`
```
public void updateConversation(Conversation conversation) {
    var selectedService = GeneralSettings.getCurrentState().getSelectedService();
    conversation.setClientCode(selectedService.getCompletionCode());
    conversation.setUpdatedOn(LocalDateTime.now());
    conversation.setModel(getModelForSelectedService(selectedService));
}
```

Issue #269, opened 2023-11-13
Add support for external llama server


p91paul:
### Describe the need of your request

Currently you allow running a local llama.cpp instance with a specified model. That's very nice, but llama.cpp has a builtin server https://github.com/ggerganov/llama.cpp/blob/master/examples/server/README.md and I'd like to connect to it over the network, without the need to start it from the CodeGPT plugin. That allows me to run the llama server on a much beefier machine than my dev PC.

### Proposed solution

Instead of just offering the option to choose model and start a server, ask for hostname and port of an existing server and connect to it, without running it locally.

hoh:
llama-cpp has a script, `api_like_OAI.py`, that should convert requests in OpenAI format to the Llama-cpp server API:
https://github.com/ggerganov/llama.cpp/blob/master/examples/server/api_like_OAI.py

I faced errors when trying to use it with CodeGPT however, and was not successful.

hoh:
See the doc on https://github.com/ggerganov/llama.cpp/tree/master/examples/server#api-like-oai

carlrobertoh:
I will improve the llama configuration in the near future, including overriding the base host, to enable support for connecting to remote servers.

exploringweirdmachines:
Also, if one wanst to use llama.cpp built with gpu support, and is already available on the system, why is CodeGPT builds his own version CPU only of llama.cpp ?

muellerti:
It would be a nice feature to add support to connect to a remote llama.cpp server like local-ai. And also also to make use of GPU Support.

hoh:
Recent versions of [llama-cpp](https://github.com/ggerganov/llama.cpp) (I can't find the exact commit) added compatibility with the OpenAI API.

This means that CodeGPT now works using a the `llama-server` command locally (or remotely) by replacing the OpenAI URL in the settings, including GPU support (in my case AMD using opencl):

![image](https://github.com/carlrobertoh/CodeGPT/assets/404665/917393e1-d1df-4fca-a888-b70af6a13bb0)

```bash
llama-server --model ./models/codellama-13b.Q4_K_M.gguf --ctx-size 2048 -ngl 35
```

PR #345, opened 2024-01-28
Allow using existing Llama Server instead of running locally

### LINKED ISSUE #260: Use CodeLLama remote server instead of running model locally

There are projects with dedicated/hosted server with already running CodeLLama remotely.

### Proposed solution

It would be great to use `hostname:port` or URL to remote CodeLLama instance instead of running it locally (too many resources needed on laptop)

### Additional context

The plugin works great so far!

PhilKes:
Fixes #260 and #269 

* Depends on: https://github.com/carlrobertoh/LLM-Client/pull/11 
* Adds option to toggle between running a local server and using an existing server:
image
image
* Allows to set `apiKey` for authentication optionally


PhilKes:
FYI: I will refactor the settings, since it gets quite messy by now

PR #349, opened 2024-02-02
Settings State + Forms refactoring

PhilKes:
I tried to investigate how to generify and unify the Tests across the different services and did quite a lot of refactoring to remove a lot of duplicates, especially in the settings and their corresponding forms. At first I wanted to investigate how easy it would be to add a new Service like Ollama, having similar settings as Llama-Cpp but instead of copy pasting everything again I tried to generify most things in common.

Main Refactorings:
* Introduce generic Setting classes that many Services have in common ( see `RemoteSettings`)
* Introduce Interfaces for `CredentialManager` since excepct `You`, all services use an `apiKey` for authentication
* Introduce generic Forms for the corresponding settings (e.g. `RemoteServiceForm`) 
* Introduce [Converters](https://plugins.jetbrains.com/docs/intellij/persisting-state-of-components.html#converter-example) for all `CompletionModel` classes


PhilKes:
@carlrobertoh I just finished the refactorings, its a lot of changes, but I think they make adding new Services a lot easier, and the code also more maintainable

src/main/java/ee/carlrobert/codegpt/settings/state/LlamaSettingsState.java
carlrobertoh, line 1: Keep in mind that these changes are not backward-compatible, meaning that after upgrading the plugin, users will lose all previously configured settings.
Previous OpenAI settings seem to get reset as well, not sure if this is a known issue.

PhilKes: Hi @carlrobertoh yes you are right, but we could implement a `ConverterProvider` which mitigates this problem, if necessary. (https://plugins.jetbrains.com/docs/intellij/persisting-state-of-components.html#migrating-persisted-values)

src/main/java/ee/carlrobert/codegpt/settings/state/OpenAISettingsState.java
```diff
@@ -5,24 +5,39 @@
...
-  private String path = BASE_PATH;
-  private String model = OpenAIChatCompletionModel.GPT_3_5.getCode();
   private boolean openAIQuotaExceeded;
 
+  public OpenAISettingsState() {
+    super("https://api.openai.com", BASE_PATH, OpenAIChatCompletionModel.GPT_3_5,
+        new OpenAICredentialsManager());
+  }
+
+  public OpenAISettingsState(String baseHost, String path, OpenAIChatCompletionModel model,
```
carlrobertoh, line 30: I'm not sure if manually creating the instance is the right way to go. Since this is an application-level service, only one instance should be created, and which can be accessed via its static `getInstance()` method.

carlrobertoh: JetBrains recommends using a [separate state class](https://plugins.jetbrains.com/docs/intellij/persisting-state-of-components.html#d264d665_40) for persistent components. Seems useful, especially when we're planning to change the entire structure anyway.

src/main/java/ee/carlrobert/codegpt/credentials/AzureCredentialsManager.java
```diff
@@ -1,70 +1,50 @@
...
 
 import com.intellij.credentialStore.CredentialAttributes;
-import com.intellij.openapi.application.ApplicationManager;
-import com.intellij.openapi.components.Service;
 import ee.carlrobert.codegpt.settings.state.AzureSettingsState;
+import java.util.Map;
+import java.util.Set;
 import org.jetbrains.annotations.Nullable;
 
-@Service
```
carlrobertoh, line 9: Hmm, what's the reason behind removing the service tag? In my opinion, these credential manages should be initialized only once, since they mostly contain logic around the Credentials Store API.

src/main/java/ee/carlrobert/codegpt/settings/state/util/CommonSettings.java
```diff
@@ -0,0 +1,31 @@
+package ee.carlrobert.codegpt.settings.state.util;
+
+import com.intellij.util.xmlb.annotations.Transient;
+import ee.carlrobert.codegpt.credentials.CredentialsManager;
+
+/**
+ * Settings that all service settings share in common.
+ */
+public class CommonSettings<T extends CredentialsManager> {
```
carlrobertoh, line 9: Can we remove this class? I don't see any reason to store the `credentialsManager` in the settings state and marking the field as transient seems just a hack to prevent the field from serialization.

src/main/java/ee/carlrobert/codegpt/settings/state/AzureSettingsState.java
```diff
@@ -5,25 +5,46 @@
...
-import ee.carlrobert.codegpt.settings.service.ServiceSelectionForm;
+import ee.carlrobert.codegpt.settings.service.AzureServiceForm;
+import ee.carlrobert.codegpt.settings.state.util.RemoteOpenAiSettings;
+import ee.carlrobert.llm.client.openai.completion.OpenAIChatCompletionModel;
 import org.jetbrains.annotations.NotNull;
 
 @State(name = "CodeGPT_AzureSettings_210", storages = @Storage("CodeGPT_AzureSettings_210.xml"))
-public class AzureSettingsState implements PersistentStateComponent<AzureSettingsState> {
+public class AzureSettingsState extends
+    RemoteOpenAiSettings<AzureCredentialsManager> implements
```
carlrobertoh, line 17: Hmm, do we really need to inherit two common params through multiple nested classes? I think this just adds an extra layer of complexity to the project and makes it harder to maintain. Also, keep in mind that `model` is now part of the Azure settings, which is unused.

A general rule of thumb is to prefer composition over inheritance.

src/main/java/ee/carlrobert/codegpt/settings/state/util/CommonSettings.java
PhilKes: I dont think using `@Transient` is hacky, the annotation is exactly for the purpose of not wanting particular fields to be serialized, which is the case here.

The `credentialsManager` is in the settings State classes because the Settings Forms also return credentials (e.g. apiKey), as you can see here:
https://github.com/carlrobertoh/CodeGPT/blob/d0dc14ef02cda5d16df2186f982b37800de6482a/src/main/java/ee/carlrobert/codegpt/settings/service/AzureServiceForm.java#L138-L151

The forms return ready-only Settings which can then simply be applied, checked for modification and also be reset in:
https://github.com/carlrobertoh/CodeGPT/blob/d0dc14ef02cda5d16df2186f982b37800de6482a/src/main/java/ee/carlrobert/codegpt/settings/state/AzureSettingsState.java#L61-L101

Removing the `credentialsManager` from the states would mean a lot copy-paste. You also have to keep in mind, that for the `LLama` Service we have 2 `CredentialsManager`s, one for the remote and one for the local config.

src/main/java/ee/carlrobert/codegpt/settings/state/OpenAISettingsState.java
PhilKes: True, I now separated the storage and state classes for every service, e.g:

https://github.com/carlrobertoh/CodeGPT/blob/c4e1f980aaa2f3330b768a7d7a53c95662d84ce5/src/main/java/ee/carlrobert/codegpt/settings/state/llama/LlamaSettingsState.java#L8-L14

src/main/java/ee/carlrobert/codegpt/settings/state/AzureSettingsState.java
PhilKes: Good thing you ask, I forgot to ask, does the Azure service not allow to select a model?
https://github.com/carlrobertoh/CodeGPT/blob/c4e1f980aaa2f3330b768a7d7a53c95662d84ce5/src/main/java/ee/carlrobert/codegpt/completions/CompletionRequestService.java#L48-L56
Does the Azure API not care about providing a model?

PhilKes: I removed the `model` from the `AzureSettingsState` and its corresponding form.

src/main/java/ee/carlrobert/codegpt/settings/state/util/CommonSettings.java
carlrobertoh: > The credentialsManager is in the settings State classes because the Settings Forms also return credentials (e.g. apiKey), as you can see here:

I understand, but why do we want to persist `credentialManager` in the state in the first place? I might be misunderstanding something here, but what value does it actually bring?

The [separate state class](https://plugins.jetbrains.com/docs/intellij/persisting-state-of-components.html#d264d665_40) approach seems to decouple the state and the settings service logic quite nicely. If we just need to access the credentials manager from each of these settings, wouldn't it be simpler to just define a single getter in the service logic?

For example:
```java
public AzureCredentialsManager getCredentialsManager() {
  return ApplicationManager.getApplication().getService(AzureCredentialsManager.class);
}
```

src/main/java/ee/carlrobert/codegpt/settings/state/AzureSettingsState.java
carlrobertoh: > Does the Azure API not care about providing a model?

Nope, I think this is configured as part of the deployment id.

> With composition over inheritance I would agree with you if its about Classes/Services that can do things, but the state classes are only POJOs with a lot of fields and getter and setter.

Yep, but `AzureSettingsState` is not a simple POJO, it's an application-level service. Idk, maybe it's just me, but even with plain POJO, it doesn't make much sense to inherit these properties. I'd rather see it as "settings **has-a** overridden/remote settings" rather than **is-a**.

src/main/java/ee/carlrobert/codegpt/settings/state/util/CommonSettings.java
PhilKes: I separated the `CredentialsManager`s from the State's, and converted every `CredentialsManager` into a Service with a `getInstance` method, for Llama I added `LlamaLocalCredentialsManager` + `LlamaRemoteCredentialsManager`. Looks cleaner now

carlrobertoh:
Why changing the UI/UX? Also, openai hyperlink seems to be gone now.

Screenshot 2024-02-05 at 01 19 16

carlrobertoh:
After the recent changes, the OpenAI key is no longer recognized.
Is it possible to break this PR into smaller units, as it's becoming too overwhelming to maintain?

PhilKes:
> After the recent changes, the OpenAI key is no longer recognized. Is it possible to break this PR into smaller units, as it's becoming too overwhelming to maintain?

With "recognized" you mean your old apiKey was not loaded properly? I just realized I changed the `CredentialAttributes` by mistake for OpenAI + Azure, I changed it to the same String key as before.
Also I dont think you can/should break this PR into smaller bits, since its all connected, State, Forms, Credentials, it would be much more effort to try to split it now, and I think it already became much cleaner.

carlrobertoh:
> With "recognized" you mean your old apiKey was not loaded properly? I just realized I changed the CredentialAttributes by mistake for OpenAI + Azure, I changed it back to the same String key as before.

Yep, I've just pulled in the latest changes and am still experiencing the same issues. The Azure credentials appear to be affected as well. After building and installing the new version of the plugin on my IDE, the keys are not being loaded correctly, despite being stored in my keychain.

src/main/java/ee/carlrobert/codegpt/credentials/manager/AzureCredentialsManager.java
```diff
@@ -21,6 +21,7 @@ public AzureCredentialsManager() {
 
   @Override
   public void apply(AzureCredentials credentials) {
+    this.credentials = credentials;
```
PhilKes, line 24: > > With "recognized" you mean your old apiKey was not loaded properly? I just realized I changed the CredentialAttributes by mistake for OpenAI + Azure, I changed it back to the same String key as before.
> 
> Yep, I've just pulled in the latest changes and am still experiencing the same issues. The Azure credentials appear to be affected as well. After building and installing the new version of the plugin on my IDE, the keys are not being loaded correctly, despite being stored in my keychain.

@carlrobertoh  I see I forgot to set the credentials not only in the PasswordSafe but also in the `CredentialsManager` too, but I dont think that has anything to do with your problem. The only lines where the `CredentialsUtil` is used is in the `CredentialsManager`s `apply` method and their constructor. I cannot reproduce your problem, since I didn't have any apiKey setup in my actual IntelliJ, only in the development build.

I will try to write some unit tests for the `CredentialsManager` and Settings

PhilKes: I added the `ServiceSettingsTest` which tests every service's `Settings` and `CredentialManager` (`apply`, `isModified`)
https://github.com/carlrobertoh/CodeGPT/blob/392cceea2f1dfb40084bb547d6f707023f4ba052/src/test/java/ee/carlrobert/codegpt/settings/state/ServiceSettingsTest.java#L43

PhilKes: Having trouble with the `CredentialsManagerTest` and `ServiceSettingsStateTest` being run in parallel, I think because they both set `ApplicationManager.setApplication`...

PhilKes: I fixed the Tests by converting the `CredentialsUtil` to a service as well. I added constructors to manually pass in `CredentialsManagers` to the `Settings` and the `PasswordSafe` to the `CredentialsUtil`. This allows me to test the `Settings` in a plain Unit Test, without using `ApplicationManager` or `BasePlatformTestCase`

src/test/java/ee/carlrobert/codegpt/settings/state/ServiceSettingsTest.java
carlrobertoh, line 1: Please use Arrange, Act and Assert (AAA) pattern when writing tests.

src/test/java/ee/carlrobert/codegpt/credentials/CredentialsManagerTest.java
```diff
@@ -0,0 +1,104 @@
...
+  private CredentialsService credentialsService;
+
+  @Before
+  public void setUp() {
+    passwordSafe = new TestPasswordSafe();
+    credentialsService = new CredentialsService(passwordSafe);
+  }
+
+  @Test
+  public void testLlamaCredentialsManager() {
```
carlrobertoh, line 30: Same thing here, please apply AAA pattern. 
Also, the method name is confusing, can't really tell what it's actually testing without reading  the implementation.

src/test/java/testsupport/TestUtil.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public static void assertCredentials(ApiKeyCredentials credentials, String apiKey) {
+    assertThat(credentials.getApiKey()).isEqualTo(apiKey);
+  }
+
+  public static void assertPassword(CommonSettings<PasswordCredentials> settings, String password) {
+    assertPassword(settings.getCredentials(), password);
+  }
+
+  public static void assertPassword(PasswordCredentials credentials, String password) {
```
carlrobertoh, line 34: Please remove this util class and use the AssertJ assertions directly where necessary.

[Favor DRY in production code, favor DAMP in test code.](https://stackoverflow.com/a/11837973)

carlrobertoh (changes requested):

carlrobertoh:
Also, could you please revert the visual UI changes for the Azure and OpenAI forms, such as the left padding of fields, the order of fields, etc.? I'm not sure why they were necessary.

src/main/java/ee/carlrobert/codegpt/completions/llama/CustomLlamaModel.java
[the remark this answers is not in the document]
PhilKes: I removed it.

src/main/java/ee/carlrobert/codegpt/util/Utils.java
```diff
@@ -0,0 +1,15 @@
+package ee.carlrobert.codegpt.util;
+
+import java.util.Objects;
+import org.jetbrains.deft.Obj;
+
+public class Utils {
```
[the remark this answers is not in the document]
PhilKes: The `com.intellij.openapi.util.text.StringUtil.equals` return `false` if a string is null and the other is empty, which would lead to `isModified` falsely returning `true` because empty TextFields always return an empty string and not `null`, but I fixed it at those places with `StringUtils.defaultString`

src/main/java/ee/carlrobert/codegpt/settings/state/azure/AzureSettingsState.java
```diff
@@ -0,0 +1,90 @@
+package ee.carlrobert.codegpt.settings.state.azure;
+
+import com.intellij.util.xmlb.annotations.Transient;
+import ee.carlrobert.codegpt.credentials.AzureCredentials;
+import ee.carlrobert.codegpt.settings.state.util.RemoteSettings;
+
+public class AzureSettingsState extends RemoteSettings<AzureCredentials> {
```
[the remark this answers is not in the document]
PhilKes: Done, `Credentials` are now only kept by `CredentialsManager`s

src/main/java/ee/carlrobert/codegpt/settings/state/llama/LlamaLocalSettings.java
```diff
@@ -0,0 +1,175 @@
...
+  private String additionalCompileParameters = "";
+
+  private PromptTemplate chatPromptTemplate = PromptTemplate.LLAMA;
+  private InfillPromptTemplate infillPromptTemplate = InfillPromptTemplate.LLAMA;
+  @OptionTag(converter = LlamaCompletionModelConverter.class)
+  protected LlamaCompletionModel model = HuggingFaceModel.CODE_LLAMA_7B_Q4;
+  private String serverPath = BUNDLED_SERVER;
+
+  public LlamaLocalSettings() {
+    this.credentials = new ApiKeyCredentials();
```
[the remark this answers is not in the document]
PhilKes: Removed `LlamaLocalCredentialsManager` + `ApiKeyCredentials` from `LlamaLocalSettings`

src/test/java/ee/carlrobert/codegpt/credentials/CredentialsManagerTest.java
PhilKes: Done

src/test/java/testsupport/TestUtil.java
PhilKes: Done

PhilKes:
> Also, could you please revert the visual UI changes for the Azure and OpenAI forms, such as the left padding of fields, the order of fields, etc.? I'm not sure why they were necessary.

Done

carlrobertoh:
What do you think about closing this PR and building the Ollama integration on top of the current master instead? If you find something important that should be included, then perhaps it can be cherry-picked into the Ollama branch.

Sorry about the massive merge conflict 🙌

PhilKes:
> What do you think about closing this PR and building the Ollama integration on top of the current master instead? If you find something important that should be included, then perhaps it can be cherry-picked into the Ollama branch directly.
> 
> I think it's not really worth resolving the conflicts unless I missed something important.
> 
> Sorry about the massive merge conflict 🙌

Why did you request all these changes if you didnt want to merge it in the first place, meanwhile doing your own refactorings? Please save me and yourself the waste of time. I dont think cherry-picking makes any sense, e.g. picking the refactoring of `LlamaSettings` but not not `LlamaServiceForm`

carlrobertoh:
By "cherry-picking," I meant selecting changes unrelated to my recent merge, such as the unit tests, llama forms refactoring, etc. For example, some of these changes can be even done independently as long as the UI component's interface is clean and concise. However, I did change some stuff related to how values are passed between the settings service and the component itself.

I'm planning to release some stuff very soon, and because of that, I'm trying to be extra cautious not to break anything, especially if it doesn't provide much feature value.

Sorry, I should have planned this better.

carlrobertoh:
@PhilKes Btw, I recently created a Discord channel that can be used for better communication, in case you're interested.
Invite link: https://discord.gg/8dTGGrwcnR

Issue #359, opened 2024-02-07
Support for together.ai OpenAPI endpoints


keriati:
### Describe the need of your request

Thank you for working on this awesome plugin. 
together.ai provides an openai compatible endpoint that could work, except they have a wide range of models with different names.

### Proposed solution

allow to use custom model name
or
load all available models from the api instead of hardcoding them.
(openai api provides a list of available models)

### Additional context

awesome plugin, love it

carlrobertoh:
Hey,

Version 2.4.0 is now publicly available, providing the option to configure custom OpenAI-compatible services.
Please let me know if it helps.

Issue #343, opened 2024-01-26
Add support for Ollama


Crown0815:
### Describe the need of your request

Run a local or remote [ollama](https://ollama.ai/library) server as backend.

### Proposed solution

Connect to the Ollama port similar to [Ollama webui](https://github.com/ollama-webui/ollama-webui)

sclark-dev:
I second this. Ollama already has a well documented API. The ability to use a local AI, and being able to choose which model would make this a must-have plugin.

gschrader:
The latest version of Ollama now has OpenAI compatibility, all is needed is to be able to specify the model name now.

gschrader:
Actually ignore my last comment the docs suggest you can copy a model with matching name:
`
ollama cp llama2 gpt-3.5-turbo
`
Seems to be working for me!

monksy:
Still it would be nice to have the plugin pull a listing of available models instead of the hard coded OpenAPI ones when you have this configured.

OleksandrWebLab:
It helped me:

https://github.com/ollama/ollama/blob/main/docs/openai.md

https://github.com/ollama/ollama/issues/2474

kalifg:
I am trying to get this to work, even with copying the model and ollama v0.1.25 — all I get from CodeGPT is "404 page not found".

PabloMzSAP:
I got this error:
404 page not found

java.lang.RuntimeException
	at ee.carlrobert.llm.completion.CompletionEventSourceListener.onFailure(CompletionEventSourceListener.java:114)
	at okhttp3.internal.sse.RealEventSource.processResponse(RealEventSource.kt:52)
	at okhttp3.internal.sse.RealEventSource.onResponse(RealEventSource.kt:46)
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:519)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)

carlrobertoh:
Version 2.4.0 is now publicly available, providing the option to configure custom OpenAI-compatible services, including the preset template for Ollama.

Please let me know if it helps.

PR #361, opened 2024-02-08
Ollama remote service

PhilKes:
https://github.com/carlrobertoh/CodeGPT/issues/343

Depends on https://github.com/carlrobertoh/llm-client/pull/14

* Adds `OllamaSettings` containing settings for using a remote Ollama Server
* Adds Chat + Code completion with `OllamaClient`
* Loads available Models from Server via `/api/tags`
* Added Ollama SVG Icons

Not in the scope of this PR:
* Trigger remote pulls of models
* Running Ollama Server locally through CodeGPT Plugin

TODOs:
* Allow other Models than `HuggingFaceModels`? Ollama allows to create customized Models with custom tags (see https://github.com/ollama/ollama/blob/main/docs/api.md#create-a-model). Unfortunately `ollama.ai/library` does not yet provide an API to query which models are available for downloading/pulling
* Correctly filter the available Models, not only by their family names but also by the available values for quantization and parmeterSize by the given `OllamaModelDetails` provided by `/api/tags`
* Refactor common methods/panels from `LlamaSettingsForm` and `OllamaSettingsForm`

Settings:
 
image


src/main/java/ee/carlrobert/codegpt/credentials/AbstractCredentialsManager.java
```diff
@@ -46,6 +46,9 @@ protected boolean isCredentialSet(String key) {
   }
 
   protected void setCredential(String key, String credential) {
+    if (credential == null) {
+      return;
+    }
     passwordSafe.setPassword(credentialMapping.get(key), credential);
     credentialCache.put(key, credential);
```
PhilKes, line 53: I got an error here, since I havent set any Azure api key and the map doesnt allow null values

PhilKes:
In the most recent commit I added support for custom `OllamaModel`s by mapping the `/api/tags` infos to `OllamaParentModel` and `OllamaChildModel` values. The Ollama API returns quantization, parametersize so we should be able to load them dynamically. We even get the file size, but unfortunately not the approx. RAM required, which is hard coded for `HuggingFaceModel`s for LLaMa, but maybe we can just omit this info for Ollama Models.

FYI here is a response of `/api/tags` with a Codellama model installed:
```json
"models": [
        {
            "name": "codellama:7b-instruct-q4_K_M",
            "model": "codellama:7b-instruct-q4_K_M",
            "modified_at": "2024-01-30T15:46:17.632920394+01:00",
            "size": 4081107848,
            "digest": "dfd9bbf35961b4480dc476bb9a32730c3ede03b18a52f1b0901e74d43d338667",
            "details": {
                "parent_model": "",
                "format": "gguf",
                "family": "llama",
                "families": null,
                "parameter_size": "7B",
                "quantization_level": "Q4_K_M"
            }
        }
    ]
```

src/main/java/ee/carlrobert/codegpt/credentials/AbstractCredentialsManager.java
carlrobertoh: Nice catch!

keriati:
Ollama added now openAI compatibility, I tested it and it works for me, however I had the create a "fake model file" that has the same name as openai chatgpt models.

I would suggest to improve the openAI API compatibility in a way that the available models are loaded from the API and the selection is not hardcoded.

Would also resolve : #359

PhilKes:
> Ollama added now openAI compatibility, I tested it and it works for me, however I had the create a "fake model file" that has the same name as openai chatgpt models.
> 
> I would suggest to improve the openAI API compatibility in a way that the available models are loaded from the API and the selection is not hardcoded.
> 
> Would also resolve : #359

@keriati thanks for the info, I didnt notice they were gonna adopt the OpenAI API. But I think the Ollama API is actually superior in terms of metadata and flexibility for using your own servers, the [/api/tags](https://github.com/ollama/ollama/blob/main/docs/api.md#list-local-models) gives much more information as the OpenAI [/v1/models](https://platform.openai.com/docs/api-reference/models/list). Ollama also provides endpoints for creating/deleting/copying models, as well as pulling Models remotely that are not yet installed/downloaded on the server yet, which we could leverage later. 
So I think for Ollama we should stick with the actual Ollama API, but for the OpenAI Service Form, you are right we should probably leverage the [/v1/models](https://platform.openai.com/docs/api-reference/models/list) to allow selecting non-hardcoded models. Unfortunately the OpenAI API really doesnt give much metadata except the models name/id, which is a bit disappointing. But we should address the OpenAI Service changes in another PR.

carlrobertoh:
In my opinion, setting up a new configuration every time someone requests for a different API provider just isn't scalable nor maintainable. I think we should have only one highly customizable service that lets you tweak everything from top to bottom. Things like prompt templates, request methods, formats, headers, etc., along with multiple pre-configured templates. Then, it really doesn't matter where you are hosting your model or which API provider you are using. 

However, having the option to simply click and start the server will be useful for users who don't really want to deal with all these configurations.

PhilKes:
> In my opinion, setting up a new configuration every time someone requests for a different API provider just isn't scalable nor maintainable. I think we should have only one highly customizable service that lets you tweak everything from top to bottom. Things like prompt templates, request methods, formats, headers, etc., along with multiple pre-configured templates. Then, it really doesn't matter where you are hosting your model or which API provider you are using.
> 
> However, having the option to simply click and start the server will be useful for users who don't really want to deal with all these configurations.

That was one reason for my refactorings...
I am also not sure what you mean with "one highly customizable service", do you mean 1 common RequestConfiguration Form? If yes you have to keep in mind that the different service use different parameters/options, that other services might not have (e.g. [Ollama completion](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-completion) vs [OpenAI completion](https://platform.openai.com/docs/api-reference/chat/create).

Also I thought about just removing the feature to run Llama locally from the IDEA, I mean I someone really wants to do that they can simply use [llama-cpp-python](https://github.com/abetlen/llama-cpp-python?tab=readme-ov-file#openai-compatible-web-server) or with ollama its even simpler to run a server locally,  so I dont think we need to bother integrating llama-cpp in this plugin. If someone runs a LLamaCpp/Ollama locally they can simply use `localhost:<PORT>` for the `baseHost`. It would focus the plugin more on what is really important, having a nice developer experience in IntelliJ with LLM support.

carlrobertoh:
> I am not sure what you mean with "one highly customizable service", do you mean 1 common RequestConfiguration Form + State? If yes you have to keep in mind that the different service use different parameters/options, that other services might not have (e.g. [Ollama completion](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-completion) vs [OpenAI completion](https://platform.openai.com/docs/api-reference/chat/create)).

Exactly, a single generic form that provides everything needed to connect a custom provider. For example, if you make the request json body configurable, then it doesn't really matter what type of parameters each service expects. It's up to the user to configure it accordingly. Of course, we can provide a "pre-configured templates" or somthing similar. This looks like a lot of work, but it definitely reduces the need to create a new provider configuration each time.

> Also I thought about just removing the feature to run Llama locally from the IDEA, I mean if someone really wants to do that they can simply use [llama-cpp-python](https://github.com/abetlen/llama-cpp-python?tab=readme-ov-file#openai-compatible-web-server) or with ollama its even simpler to run a server locally, so I dont think we need to bother integrating llama-cpp in this plugin. If someone runs a LLamaCpp/Ollama locally they can simply use localhost:<PORT> for the baseHost. It would focus the plugin more on what is really important, having a nice developer experience in IntelliJ with LLM support.

As I mentioned earlier, having the option to easily start the server is useful for users who are not interested in configuring and running any LLMs themselves but would still like to use the extension's functionality with local models.

SupremeVoid:
Any progress on this?

PhilKes:
> Any progress on this?

I did not yet continue the PR, since @carlrobertoh suggested a radically different approach which would take a lot more time. I am currently looking at ways to implement such a generic `llm-client`, but it would definetely take much more time before being able to implement and polish it

PR #371, opened 2024-02-18
feat: Use /v1/chat/completions for OpenAI code-completions

### LINKED ISSUE #369: Code Inline Completion change to v1/chat/completions
### What happened?

I am using a dedicated service provider for OpenAI models and use `Base host`. In a new function (Code Inline Completion), the `v1/completions` method is called. This method is not available from my service provider. This method is Legacy in the OpenAI documentation, and also it is most often not available on new models. 

Is it possible to change the method to `v1/chat/completions`?

### CodeGPT version

2.3.0

PhilKes:
Fixes #369.

This PR changes the Code-Completion for `OpenAI` ServiceType to use `/v1/chat/completions` instead of `/v1/completions`.
There is no `suffix` parameter for OpenAI Chat-Completions, therefore the FIM prompt template is used instead as a single "user" message.

Unfortunately the results dont look right to me just yet, I'm not sure why, here's an example: 
image

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/codecompletions/CodeCompletionRequestProvider.java
+++ b/src/main/java/ee/carlrobert/codegpt/codecompletions/CodeCompletionRequestProvider.java
@@ -2,8 +2,12 @@ package ee.carlrobert.codegpt.codecompletions;
 
 import ee.carlrobert.codegpt.completions.llama.LlamaModel;
 import ee.carlrobert.codegpt.settings.service.llama.LlamaSettings;
+import ee.carlrobert.codegpt.settings.service.openai.OpenAISettings;
+import ee.carlrobert.codegpt.settings.service.openai.OpenAISettingsState;
 import ee.carlrobert.llm.client.llama.completion.LlamaCompletionRequest;
-import ee.carlrobert.llm.client.openai.completion.request.OpenAITextCompletionRequest;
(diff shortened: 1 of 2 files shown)


reneleonhardt:
@PhilKes Thank you for providing a fix, I rebased for you and converted your code to Kotlin.
`OpenAISettingsState` doesn't contain isUsingCustomPath and path anymore, I commented it out.

Can you check if the current code fixes the problem you have seen? 
https://github.com/reneleonhardt/CodeGPT/tree/openai-code-use-chat-completions

PhilKes:
> @PhilKes Thank you for providing an implementation, I rebased for you and converted your code to Kotlin. `OpenAISettingsState` doesn't contain `isUsingCustomPath` and `path` anymore, I commented it out.
> 
> Can you check if the current code fixes the problem you have seen? https://github.com/reneleonhardt/CodeGPT/tree/openai-code-use-chat-completions

I am not sure if this PR is still relevant, I haven't been working on CodeGPT for some weeks now.
See https://github.com/carlrobertoh/CodeGPT/issues/369#issuecomment-1951444495
Perhaps this can be closed @carlrobertoh ?

adam6806:
Can I ask why "Code completions are only available with gpt-3.5-turbo-instruct model"? This model is much more expensive than say 4o-mini. Is it not possible to get code completions to work with that model?

PR #378, opened 2024-02-20
Enable remote server settings for Windows + Mixtral Instruct template

young-developer:
Enabled remote server for Windows:
image

- Mixtral Instruct template was added.


carlrobertoh:
Many thanks! 👍

PR #399, opened 2024-03-07
Implement support for You Pro modes

melan:
It adds support for several new search modes: Default, Agent, Custom with various 3rd party models and Research

Depends on lllm-client changes here: https://github.com/carlrobertoh/llm-client/pull/16

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/settings/service/you/YouSettings.java
+++ b/src/main/java/ee/carlrobert/codegpt/settings/service/you/YouSettings.java
@@ -4,6 +4,7 @@ import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.components.PersistentStateComponent;
 import com.intellij.openapi.components.State;
 import com.intellij.openapi.components.Storage;
+import ee.carlrobert.llm.client.you.completion.YouCompletionMode;
 import org.jetbrains.annotations.NotNull;
 
 @State(name = "CodeGPT_YouSettings", storages = @Storage("CodeGPT_YouSettings.xml"))
@@ -20,6 +21,9 @@ public class YouSettings implements PersistentStateComponent<YouSettingsState> {
   @Override
   public void loadState(@NotNull YouSettingsState state) {
     this.state = state;
+    if (this.state.getChatMode() == null) {
+      this.state.setChatMode(YouCompletionMode.DEFAULT);
+    }
   }
 
   public static YouSettingsState getCurrentState() {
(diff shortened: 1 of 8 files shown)


melan:
I don't include these two changes to gradle.kts: 
* add okhttp3 libraries - it didn't build the code without me adding them
* switch to a local copy of the llm-client

src/main/java/ee/carlrobert/codegpt/toolwindow/chat/ui/textarea/ModelComboBoxAction.java
```diff
@@ -87,8 +94,16 @@ public JComponent createCustomComponent(
         getLlamaCppPresentationText(),
         Icons.Llama,
         presentation));
-    actionGroup.addSeparator();
-    actionGroup.add(createModelAction(ServiceType.YOU, "You.com", Icons.YouSmall, presentation));
+    actionGroup.addSeparator("You.com");
+    List.of(
```
carlrobertoh, line 98: Could we render these new modes/models only when the user has an active You.com session? Otherwise, they will occupy too much space in the popup.

melan: They blocked by account while I tested it by logging in and out, but done 🤣

src/main/java/ee/carlrobert/codegpt/settings/service/you/YouSettings.java
```diff
@@ -20,6 +21,9 @@ public YouSettingsState getState() {
   @Override
   public void loadState(@NotNull YouSettingsState state) {
     this.state = state;
+    if (this.state.getChatMode() == null) {
+      this.state.setChatMode(YouCompletionMode.DEFAULT);
```
carlrobertoh, line 25: I think this is unnecessary, you can set it already as the default value. It shouldn't cause any problems since it's a new field.

src/main/java/ee/carlrobert/codegpt/toolwindow/chat/ui/textarea/ModelComboBoxAction.java
carlrobertoh: I had the same problem during the initial implementation 😄

carlrobertoh (approved):

src/main/java/ee/carlrobert/codegpt/settings/service/you/YouSettings.java
melan: I'm not a java developer, sorry. Can you give me an example of how to do this?

ChatMode and Model aren't on the setup form, so they aren't stored in the xml, but the mode should be initialized into something not-null. that's why this code is here

carlrobertoh:
```java
private boolean useGPT4Model;
private YouCompletionMode chatMode = YouCompletionMode.DEFAULT;
private YouCompletionCustomModel customModel;
```

https://github.com/carlrobertoh/CodeGPT/pull/399/files#diff-def42ea7a06eeb949bd57cd4152c41ab0d8b887c8d95fac817a95130f07d63d0R12

PR #424, opened 2024-03-26
feat: streaming code completion support

wsxqyws:
This PR switch code completion to streaming mode.

Please feel free to modify these code if needs.

Here is the demo. The first completion is performered without interruption. The next is early stopped by shortcuts.

https://github.com/carlrobertoh/CodeGPT/assets/10071320/97044e80-48f7-4980-8623-fe4349aeef97

Here are the main changes:
1. switch code completion to streaming, it reduces user's waiting time and looks pretty cool
2. Add `Ctrl + Alt + i` shortcut to trigger completion manually with no condition
3. Add `Ctrl + n` to accept n lines of compeltion result. (`n in [1, 2, 3, 4, 5]`)
4. Add `Max tokens for code completion` configurable (`default = 64`)
5. User can `early stop` code completion by shortcuts ( `TAB` or `Ctrl + n`)

Note: I didn't modify the previous test code for the autocompletion logic so it failed. The reason for keeping it is because this code may be used in the future.


wsxqyws:
Hello,

I have noticed the merge record on your end, and admittedly, the timing was not quite ideal. When I submitted this PR, I had some reservations because using a streaming approach for code completion is quite uncommon. If you also find the streaming approach to completion beneficial and it won't interfere with your future plans, I will make an effort to implement this functionality under the new API and submit it then.

I will draft this PR for now and open a new one at a more appropriate time. 
Also, you have been doing an amazing job; CodeGPT is extremely useful!

PR #425, opened 2024-03-26
#422 - Migrate to a new inline completion api

carlrobertoh:


src/main/kotlin/ee/carlrobert/codegpt/codecompletions/CodeGPTInlineCompletionProvider.kt
```diff
@@ -0,0 +1,104 @@
...
+            try {
+                send(InlineCompletionGrayTextElement(it))
+            } catch (e: Exception) {
+                LOG.error("Failed to send inline completion suggestion", e)
+            }
+        }
+    }
+
+    override fun isEnabled(event: InlineCompletionEvent): Boolean {
+        return ConfigurationSettings.getCurrentState().isCodeCompletionsEnabled
```
kkarnauk, line 73: Hello! I'm one of the developers of this Inline Completion API. Thank you very much for migrating to this API. 

I'd like to be sure that you are aware that there can be custom InlineCompletionEvent, and you should not say `isEnabled = true` to all of them. Some events can be internal implementation of a provider, so reacting to all the events may break some plugin.

The default events are: `DirectCall`, `aDocumentChange` and `LookupChange`.

carlrobertoh: Ohh, thank you for pointing that out! Listening for `DocumentChange` events is actually all I need at the moment.

While I have you here, what would be the best or recommended way of implementing partial acceptance? For example, besides accepting the entire response by hitting the TAB key, I would like to accept one word at a time by moving the cursor to the right with the `CMD + →` command. Is this even possible with the current implementation?

kkarnauk: Yes, it's possible with the new version of API in 2024.1. There is `InlineCompletionSuggestionUpdateManager` that can be registered in your provider. However, for that you need to make a document change that invalidates the completion (any document change except typings leads to hiding of inline completion). You can wrap your code that inserts a part of suggestion in `InlineCompletionHandler.withIgnoringDocumentChanges` and then it'll work. But this method is still experimental and can be dropped at some point.

We understand that it's an important part of the API and at some point there will be a stable way to do it. Then we'll mark the method mentioned above as deprecated.

kkarnauk: JFYI, the release of 2024.1 is going to be soon.

carlrobertoh: Super! I will play around with the API a bit more and try to implement the desired behaviour. I will get back to you in case of questions.

Many thanks again! 💯

carlrobertoh: @kkarnauk I was able to implement the partial inlay acceptance following your suggestions. Everything seems to be working fine, except for the partial multi-line acceptance. Somehow, the suggestion gets invalidated once I try to accept a partial word on the next line. I'm not sure if this is even supported, or if it's a bug on my end. I haven't had the chance to debug it yet.

I'd really appreciate it if you could take a quick peek at the implementation to see if I'm on the right track.
https://github.com/carlrobertoh/CodeGPT/compare/platform/2024.1?expand=1

Thank you!

kkarnauk: I've quickly reviewed your implementation. Looks like the root of the problem is that you insert a suggestion's part inside an action. You should do all the work inside `CodeCompletionSuggestionUpdateAdapter.onCustom`.

Also, could you please show a video where your implementation doesn't work?

kkarnauk: Oh, I think I found the problem. The `startOffset` and `endOffset` in `InlineCompletionRequest` are responsible for the text range that was affected to cause the completion. It should be `(the current caret offset, the current caret offset + the inserted text length)`. You should not use the offsets from the previous session, please use `editor.caretModel.currentOffset`.

carlrobertoh: Thanks! It seems that, in addition to the incorrect offset, the order of execution was also incorrect. The session was updated only after the caret change, which caused the inlays to disappear due to a failed condition in `InlineSessionWiseCaretListener`.

kkarnauk: Hey! Have you succeeded in getting the desired behavior?

carlrobertoh: Hey! Yes, it seems to be working now.
Thanks!

PR #433, opened 2024-04-04
fix: correctly handle changed files to generate a commit message #338

qoter:
1. Change the method we use to retrieve files that were changed and included in the commit. The old method did not work as mentioned in issue #338. I followed the advice provided [here](https://intellij-support.jetbrains.com/hc/en-us/community/posts/10709161728018/comments/10729508917266).

2. Remove the "--cached" option when retrieving the `git diff` because the Idea does not perform a `git add` when a file is selected. As a result, with the cached option, we encounter an empty diff.

3. For unversioned files, add the file content to the prompt as well since there is no way to force `git diff` to show these files.


carlrobertoh:
Unfortunately, when removing the `--cached` flag, the git diff will not be in sync anymore. For example, I do a lot of staging before the actual commit, and now when I try to get the diff, the cached/staged files aren't included in it.

I think we should probably fetch both diffs, with `--cached` and without, and include both of them in the final prompt.

PR #447, opened 2024-04-11
chore: Convert Java tests to Kotlin

reneleonhardt:
Let's start with test code conversion first 😄


reneleonhardt:
@carlrobertoh Git is a content tracker, unfortunately renames/moves are not recognized if the threshold of changes are too high. Makes converting not easier... 😅

carlrobertoh:
Thanks! I didn't really expect you to convert everything at once, but instead to do it piecemeal as you go through the old code. 🙂

reneleonhardt:
I can imagine, but in that case I would expect an hourly rate 😋
There can not as much go wrong if tests are converted first before production code, now we can observe test runs if there is any different behavior.

Issue #448, opened 2024-04-11
CodeGPT suppresses the new Jetbrains feature for local Full Line Code Completion, even if it does not work with current configuration


niceapps-ch:
### What happened?

CodeGPT does not support code completion when connected to an OpenAI Service or local via Ollama or LM Studio.
JetBrains introduced an integrated full line code completion feature that uses a very small local model. This new feature does not work when CodeGPT is active, even if CodeGPT does not do any code completion with the active configuration.

Proposal:
- Make code completion configurable, so I can turn it off to use the native feature of IntelliJ and still use all the rest of CodeGPT.

### CodeGPT version

2.6.0

### Operating System

macOS

boswelja:
Mildly unrelated, but how well does the built-in code completion work? I use Android Studio, so I won't get to use it for a little bit

carlrobertoh:
Thanks for reporting, I'll take a look. 
This should only happen in IDE versions starting from 2024.1.

niceapps-ch:
> Mildly unrelated, but how well does the built-in code completion work? I use Android Studio, so I won't get to use it for a little bit

I'm not sure, didn't get to try it productively yet. And probably won't until this is fixed, because I'm too happy with CodeGPT :)

reneleonhardt:
The bug was introduced in 2.6.0 https://github.com/carlrobertoh/CodeGPT/commit/f0172722c75ae50d2ea895f68cbef0c90bbbcc7f#diff-79d11c7104a03e0a8b9c00e9e8cc953346a515a4feeb91fb98334122664f4a73R47
I will provide a fix later and write a test when there is time, pleasy try again with the next release.

carlrobertoh:
Most likely, CodeGPT's inline suggestions provider is interfering with IntelliJ's one, altho I'm not 100% sure.
https://github.com/carlrobertoh/CodeGPT/blob/master/src/main/resources/META-INF/plugin.xml#L43

reneleonhardt:
@niceapps-ch Can you try again after disabling the plugin `Full line code completion` or disabling the setting?
https://blog.jetbrains.com/blog/2024/04/04/full-line-code-completion-in-jetbrains-ides-all-you-need-to-know/#how-to-tweak-the-feature

carlrobertoh:
@niceapps-ch which IDE are you using, and how did you confirm that the native completions are suppressed? 
[...]

PR #450, opened 2024-04-11
Draft: fix: Use platform inline completion if custom is disabled (#448)

reneleonhardt:
No change yet, just tests for current behavior.


carlrobertoh:
Closing this since the problem has been resolved.

Issue #446, opened 2024-04-11
High CPU usage with v2.6.0


fishermans:
### What happened?

After updating from 2.5.1 to 2.6.0 on Intellij IDEA Ultimate 2024.1, the CPU usage is high even on idle. The fans are running on highest speed all the time on my MacBook Pro 2020 Intel. This is also happening when I am simply launching IntelliJ and opening any of my projects without further actions. Cache cleanup did not help.

All is working fine after I downgraded to 2.5.1 again.

Any suggestions?

### CodeGPT version

2.6.0

### Operating System

macOS

carlrobertoh:
Hey, without a relevant thread dump, it's hard to say. One possible option is that the file watcher CodeGPT uses to check for new screenshots is taking a toll on your CPU. Try disabling the feature and see if it makes any difference.

<img width="613" alt="Screenshot 2024-04-11 at 11 50 37" src="https://github.com/carlrobertoh/CodeGPT/assets/15891408/6e2f27a2-240d-45c0-87e0-3ba64f104e85">

fishermans:
Thank you very much. It seems to work now.

BTW: I was not able to disable this setting. The "Apply" button remained grayed out until I unchecked at least one of the other options as well. So I had to disable screenshot and method name lookup suggestions to be able to apply it. I then re-enabled the method name option again.

For testing purposes, I removed all CodeGPT settings files from the options folder and tried again with the default settings, but this behavior persisted.

carlrobertoh:
I'm glad you got it working!
I'll keep the ticket open until I find a fix for it and in case anyone else runs into the same issue.

reneleonhardt:
> BTW: I was not able to disable this setting. The "Apply" button remained grayed out until I unchecked at least one of the other options as well. So I had to disable screenshot and method name lookup suggestions to be able to apply it. I then re-enabled the method name option again.
> 
> For testing purposes, I removed all CodeGPT settings files from the options folder and tried again with the default settings, but this behavior persisted.

@fishermans There was a bug in the settings, thank you for reporting so I could fix it, please try again in the next release after 2.6.0.

carlrobertoh:
This issue should be fixed in the new version 2.6.1. Please reopen the ticket if it is still reproducible.

kaorukobo:
@carlrobertoh Thanks for trying to fix this.

[...]

PR #451, opened 2024-04-11
fix: Use correct setting for updates and screenshot checkboxes (#446)

reneleonhardt:
Settings for Updates and Screenshot checkboxes were mixed up internally


carlrobertoh:
Ouch

Issue #411, opened 2024-03-14
Api token resets after restart PhpStorm client


timoheutsrs:
### What happened?

Every time, the PHPstorm client (our organization works with the Jetbains gateway) restarts, the API token must be entered again.

### Steps to reproduce

- Go via Jetbrains gateway to PHPStorm Client
- Ask question to CodeGPT, notice message API token not provided
- Add token in settings
- Ask the same question again, an answer will be given
- Restart the PhpStorm client
- Aks a question, notice that again the Api Token must be provided

### CodeGPT version

2.5.0

### Operating System

Linux

SamuelBohumel:
Hello, same problem here

reneleonhardt:
@timoheutsrs @SamuelBohumel Thank you for reporting, can you tell us what services you selected and configured with an API token?

timoheutsrs:
[cid:06982262-9407-4fb6-be79-54477aa248c8]
This is my config, ofcourse I readded the api key for now. But after each restart of the phpStorm api client the key is cleared

Met vriendelijke groet,
​
TimoHeuts
Developer
***@***.***
+31 6 236 594 53
________________________________
Van: Rene Leonhardt ***@***.***>
Verzonden: vrijdag 12 april 2024 20:31
Aan: carlrobertoh/CodeGPT ***@***.***>
CC: Timo Heuts | Retail Solutions ***@***.***>; Mention ***@***.***>
Onderwerp: Re: [carlrobertoh/CodeGPT] Api token resets after restart PhpStorm client (Issue #411)


@timoheutsrs<https://github.com/timoheutsrs> @SamuelBohumel<https://github.com/SamuelBohumel> Thank you for reporting, can you tell us what services you selected and configured with an API token?

carlrobertoh:
It should be fixed in the latest version. Please reopen the ticket if it's still reproducible.

timoheutsrs:
Yes its fixed, thanks! Appreciate it!

Met vriendelijke groet,
​
TimoHeuts
Developer
***@***.***
+31 6 236 594 53
________________________________
Van: Carl-Robert ***@***.***>
Verzonden: dinsdag 23 april 2024 13:32
Aan: carlrobertoh/CodeGPT ***@***.***>
CC: Timo Heuts | Retail Solutions ***@***.***>; Mention ***@***.***>
Onderwerp: Re: [carlrobertoh/CodeGPT] Api token resets after restart PhpStorm client (Issue #411)


[...]

PR #456, opened 2024-04-13
fix: Persist credentials (#411)

reneleonhardt:
Persist credentials, not only settings state:
* Credentials in form are filled/reset from password store
* Credentials are persisted to store on settings apply/ok (actual fix)
* Streamlined credentials to be null if user entered blank (whitespace)
* Credentials form and persistence is tested

Gotchas:
* YouSettingsForm persists to store directly, I suggest to move that to `GeneralSettingsConfigurable` like all other services do (user clicks Apply/OK)

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/completions/CompletionClientProvider.java
+++ b/src/main/java/ee/carlrobert/codegpt/completions/CompletionClientProvider.java
@@ -1,8 +1,9 @@
 package ee.carlrobert.codegpt.completions;
 
+import static ee.carlrobert.codegpt.credentials.Credential.getCredential;
+
 import ee.carlrobert.codegpt.CodeGPTPlugin;
 import ee.carlrobert.codegpt.completions.you.YouUserManager;
-import ee.carlrobert.codegpt.credentials.CredentialsStore;
 import ee.carlrobert.codegpt.credentials.CredentialsStore.CredentialKey;
 import ee.carlrobert.codegpt.settings.advanced.AdvancedSettings;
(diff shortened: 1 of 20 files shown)


carlrobertoh:
> - Credentials added to settings state

This poses a security risk because the sensitive data will now be stored in plain text alongside other configuration files.

reneleonhardt:
> This poses a security risk because the sensitive data will now be stored in plain text alongside other configuration files.

Now I found the culprit, the password store was only used for loading credentials, but not actually storing them 😅
I would still suggest to streamline You password handling with all other services.

reneleonhardt:
I moved You password persistence from Sign In to Apply like all others, and fixed modelChanged detection.
Did you enable auto-squashing in this repo, or should I squash manually?

carlrobertoh:
Yeah, originally, I loaded the credentials lazily whenever they were required. However, the issue arises because the `GeneralSettings` configuration form is bound to all the services. Thus, whenever there is a change in one of the services, all fields, including the credentials, need to be re-validated, and this causes an error/warning because the process runs on the EDT.

However, loading all credentials at once introduces another problem, because you must grant access to each of these credentials as soon as you start the IDE.

carlrobertoh:
The UI is the bottleneck. All these services/providers need to have their own configuration form and must implement the `com.intellij.openapi.options.Configurable` interface.

So, essentially, we need to revert back to the original lazily loaded logic and change the way how the services are configured.

Screenshot 2024-04-15 at 13 31 46

PR #460, opened 2024-04-14
Personas Feature

booka66:
Hello Carl!

We've built a VS Code extension, byoLAD (bring your own LLM Augmented Development; see: https://marketplace.visualstudio.com/items?itemName=BeanLabBYU.vscode-byolad), which has similar functionality to your CodeGPT JetBrains extension. We're impressed with the work you've done and would love to potentially contribute to CodeGPT.

One of the main appeals of both of our extensions is the ability to choose which LLM and provider to use. With byoLAD, we wanted to take customizability one step further by allowing users to easily configure system prompts associated with specific models/providers. We call these configurations "Personas", and they enable users to switch LLM modality for specific or general use cases.

For example, we've created a "Rubber Duck" persona aimed towards teaching and understanding rather than immediate code completion. This has proved useful for new CS students, but it is also useful for experienced coders as well. We've also found personas useful for tailoring LLM responses to specific languages, libraries, and conventions used in our projects. These types of personas allow us to avoid unnecessary responses which include irrelevant code. If you're curious, I've included a quick video displaying a few of these personas (see: https://drive.google.com/file/d/1dayaGir4fqAwi3RKfrnYf0x9tVx6lelR/view?usp=drive_link).

We're interested in expanding our VS Code extension to JetBrains, which led us to CodeGPT. In an attempt to explore how much of an effort it would take to introduce personas to the extension, I forked the CodeGPT repository. I found it easily enough to implement the full feature and I think the changes might be ready.

We realize focusing on user configuration around personas instead of individual conversations is a little bit of a paradigm shift for CodeGPT, so we're interested to hear what you think of the idea and how you would feel about having the changes in CodeGPT.

I'd love to hear what you think!


carlrobertoh:
Hi Jacob,

Thank you for considering CodeGPT!

This is something I have been thinking about as well, especially since I use CodeGPT's chat interface for non-technical stuff quite a bit. However, instead of writing these prompts one by one, I was considering exporting them in bulk (e.g., [awesome-chatgpt-prompts](https://github.com/f/awesome-chatgpt-prompts)) or adding an integration with some well-known prompt library, if one exists.

I haven't had the chance to look at your changes yet, but given the upcoming improvements I want to make around the configuration settings UI, I would like to put this feature on hold for at least a week or so.

I like the idea, though, but it needs to be discreet and not draw too much attention away from the actual purpose behind the plugin.

carlrobertoh:
Hey! I'm closing this PR since I've implemented this in another branch. I've also added the Rubber Duck persona to the list. 

If you have any other cool ideas about what could be added, then I'm all ears 🙂 

Link to the new PR - https://github.com/carlrobertoh/CodeGPT/pull/638

PR #461, opened 2024-04-15
Stream code completion and cancel early on newline

ChuangLee:
- Stream completion results character by character to provide a more responsive user experience
- Cancel the completion request early when a newline character is encountered to save token usage costs
These changes improve performance, reduce costs, and enhance the overall user experience. Please review and provide feedback.

The current completion suggestion implementation waits for all results before returning the content before the first newline character. This approach is slow and wastes token usage.
```
            val output =
                if (messageBuilder.contains("\n"))
                    messageBuilder.substring(0, messageBuilder.indexOf("\n"))
                else messageBuilder.toString()
            completed(output)
```

By the way, I am a colleague of wsxqyws and I am taking over his work to implement the streaming completion feature for CodeGPT using the new API.

### DIFF
--- a/src/main/kotlin/ee/carlrobert/codegpt/codecompletions/CodeGPTInlineCompletionProvider.kt
+++ b/src/main/kotlin/ee/carlrobert/codegpt/codecompletions/CodeGPTInlineCompletionProvider.kt
@@ -43,14 +43,11 @@ class CodeGPTInlineCompletionProvider : InlineCompletionProvider {
             currentCall.set(
                 CompletionRequestService.getInstance().getCodeCompletionAsync(
                     infillRequest,
-                    CodeCompletionEventListener(infillRequest) { message, cancel ->
-                        if (cancel) {
-                            cancelCurrentCall()
-                        }
-                        request.editor.putUserData(CodeGPTKeys.PREVIOUS_INLAY_TEXT, message)
+                    CodeCompletionEventListener(infillRequest) {
+                        request.editor.putUserData(CodeGPTKeys.PREVIOUS_INLAY_TEXT, it)
                         launch {
                             try {
-                                trySend(InlineCompletionGrayTextElement(message))
+                                trySend(InlineCompletionGrayTextElement(it))
                             } catch (e: Exception) {
                                 LOG.error("Failed to send inline completion suggestion", e)
                             }
@@ -78,7 +75,7 @@ class CodeGPTInlineCompletionProvider : InlineCompletionProvider {
 
     class CodeCompletionEventListener(
         private val requestDetails: InfillRequestDetails,
-        private val completed: (String, Boolean) -> Unit
+        private val completed: (String) -> Unit
     ) : CompletionEventListener<String> {
 
         private var isStreaming = false
@@ -86,8 +83,10 @@ class CodeGPTInlineCompletionProvider : InlineCompletionProvider {
         override fun onMessage(message: String?, eventSource: EventSource?) {
             if (message != null) {
                 isStreaming = true
-                val needCancel = message.contains('\n')
-                completed(message.takeWhile { it != '\n' }, needCancel)
+                completed(message.takeWhile { it != '\n' })
+                if (message.contains('\n') && eventSource != null) {
+                    eventSource.cancel()
+                }
             }
         }
 
@@ -102,7 +101,7 @@ class CodeGPTInlineCompletionProvider : InlineCompletionProvider {
                 )*/
             if (!isStreaming) {
                 val output = messageBuilder.toString().takeWhile { it != '\n' }
-                completed(output, false)
+                completed(output)
             }
         }
     }
--- a/src/main/kotlin/ee/carlrobert/codegpt/codecompletions/CodeGPTInlineCompletionProvider.kt
+++ b/src/main/kotlin/ee/carlrobert/codegpt/codecompletions/CodeGPTInlineCompletionProvider.kt
@@ -43,14 +43,14 @@ class CodeGPTInlineCompletionProvider : InlineCompletionProvider {
             currentCall.set(
                 CompletionRequestService.getInstance().getCodeCompletionAsync(
                     infillRequest,
-                    CodeCompletionEventListener(infillRequest) { suggestion, needCancel ->
-                        if (needCancel) {
+                    CodeCompletionEventListener(infillRequest) { message, cancel ->
+                        if (cancel) {
                             cancelCurrentCall()
                         }
-                        request.editor.putUserData(CodeGPTKeys.PREVIOUS_INLAY_TEXT, suggestion)
+                        request.editor.putUserData(CodeGPTKeys.PREVIOUS_INLAY_TEXT, message)
                         launch {
                             try {
-                                trySend(InlineCompletionGrayTextElement(suggestion))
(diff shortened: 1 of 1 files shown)


src/main/kotlin/ee/carlrobert/codegpt/codecompletions/CodeGPTInlineCompletionProvider.kt
```diff
@@ -43,11 +43,14 @@ class CodeGPTInlineCompletionProvider : InlineCompletionProvider {
             currentCall.set(
                 CompletionRequestService.getInstance().getCodeCompletionAsync(
                     infillRequest,
-                    CodeCompletionEventListener(infillRequest) {
-                        request.editor.putUserData(CodeGPTKeys.PREVIOUS_INLAY_TEXT, it)
+                    CodeCompletionEventListener(infillRequest) { suggestion, needCancel ->
```
reneleonhardt, line 46: Wouldn't `message, cancel` be more self-explaining than `suggestion, needCancel`?
```suggestion
                    CodeCompletionEventListener(infillRequest) { message, cancel ->
```

ChuangLee: Yeah! fixed.

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/CodeGPTInlineCompletionProvider.kt
```diff
@@ -75,9 +78,19 @@ class CodeGPTInlineCompletionProvider : InlineCompletionProvider {
...
+        private val completed: (String, Boolean) -> Unit
     ) : CompletionEventListener<String> {
 
+        private var isStreaming = false
+
+        override fun onMessage(message: String?, eventSource: EventSource?) {
+            if (message != null) {
+                isStreaming = true
+                val needCancel = message.contains('\n')
+                completed(message.takeWhile { it != '\n' }, needCancel)
```
carlrobertoh, line 90: This doesn't actually cancel the request (and save tokens), rather it just cancels early from the callback.

Would something like this work:

```kt
private val messageBuilder = StringBuilder()

override fun onMessage(message: String?, eventSource: EventSource?) {
    messageBuilder.append(message)
    if (messageBuilder.contains('\n')) {
        eventSource?.cancel()
    }
}
```

now, once `cancel()` is called, the `onComplete` callback should be triggered without requiring any other changes

carlrobertoh: you could also trim the result down to the first element before the newline as well

ChuangLee: But this has been verified through my testing and the Stream was indeed cancelled early. The llama server local immediately stopped the inferencing.
Below are the logs I printed in the onMessage() and cancelCurrentCall() functions.

```
[OkHttp http://localhost:8080/...] INFO ee.carlrobert.llm.completion.CompletionEventSourceListener - Request opened.
2024-04-16 08:54:39,755 [  75441]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage:public
2024-04-16 08:54:39,795 [  75481]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage: static
2024-04-16 08:54:39,834 [  75520]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage: void
2024-04-16 08:54:39,877 [  75563]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage: main
2024-04-16 08:54:39,919 [  75605]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage:(
2024-04-16 08:54:39,960 [  75646]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage:String
2024-04-16 08:54:40,001 [  75687]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage:[]
2024-04-16 08:54:40,041 [  75727]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage: args
2024-04-16 08:54:40,082 [  75768]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage:)
2024-04-16 08:54:40,123 [  75809]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage: {
2024-04-16 08:54:40,163 [  75849]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - onMessage:

2024-04-16 08:54:40,164 [  75850]   WARN - #ee.carlrobert.codegpt.codecompletions.CodeGPTInlineCompletionProvider - cancelCurrentCall called!
[OkHttp http://localhost:8080/...] INFO ee.carlrobert.llm.completion.CompletionEventSourceListener - Stream was cancelled
```

carlrobertoh: Ah right, `cancelCurrentCall` is invoked based on the `cancel` condition. Nevertheless, I would still like to see the cancelling being done inside the `onMessage` callback, then you don't need to declare any redundant class variables etc.

ChuangLee: I think you are right, updated and verified. 
and I think no action needed in onCancelled(), Code is more concise now.

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/CodeGPTInlineCompletionProvider.kt
```diff
@@ -78,6 +78,18 @@ class CodeGPTInlineCompletionProvider : InlineCompletionProvider {
         private val completed: (String) -> Unit
     ) : CompletionEventListener<String> {
 
+        private var isStreaming = false
+
+        override fun onMessage(message: String?, eventSource: EventSource?) {
+            if (message != null) {
+                isStreaming = true
```
carlrobertoh, line 85: Can we get rid of this class variable?

This should be enough:
```kt
if (message.contains('\n') && eventSource != null) {
    eventSource.cancel()
}
```

ChuangLee: isStreaming and `completed(message.takeWhile { it != '\n' })` is for streaming completion to reduce user waiting anxiety
I think we can remove isStreaming variables and onComplete() method assuming all LLM services support streaming responses

ChuangLee: isStreaming and onComplete() have been removed

Issue #457, opened 2024-04-13
Config not saving API key


loevstroem:
### What happened?

When you set the API key for the Anthropic Service under Settings | Tools | CodeGPT and then restart PHPStorm, the API key resets to null. This suggests it's either not saving correctly or resetting upon reload.

### CodeGPT version

2.6.0-241

### Operating System

macOS

QuadStingray:
Same on Custom Open Ai Configuration. 

Workaround is to change variable for authentication on header from variable to your api key.

reneleonhardt:
I hope my contribution https://github.com/carlrobertoh/CodeGPT/pull/456 will be merged, it should solve all non-persisted API key problems 😄
Kudos to @QuadStingray for finding a workaround in the mean time, at least for Custom OpenAI 👍

kriit24:
same thing in 2026

PR #465, opened 2024-04-17
Persist credential changes into the PasswordSafe

sisve:
I believe this fixes the currently reported issues where credentials are not persisted between application restarts.

* https://github.com/carlrobertoh/CodeGPT/issues/457
* https://github.com/carlrobertoh/CodeGPT/issues/411


carlrobertoh:
Hi, I will start preparing for a new release today. You can expect it to be available within the next 1-2 days.

PR #475, opened 2024-04-18
feat: add include file in context to editor context menu

### LINKED ISSUE #428: Read file

Hi author, is it possible to add a function that reads a file or references a file?
Optimised token formatting to recognise replicated IDEA token formats.

### Proposed solution

_No response_

### Additional context

_No response_

PhilKes:
This PR closes #428 

As suggested in https://github.com/carlrobertoh/CodeGPT/issues/428#issuecomment-2030303424 this adds the `Include in Context...` action in the Editor's context-menu.


src/main/java/ee/carlrobert/codegpt/actions/IncludeFilesInContextAction.java
```diff
@@ -93,11 +93,6 @@ public void nodeStateChanged(@NotNull CheckedTreeNode node) {
   }
 
   private @Nullable FileCheckboxTree getCheckboxTree(DataContext dataContext) {
-    var psiElement = CommonDataKeys.PSI_ELEMENT.getData(dataContext);
```
PhilKes, line 96: I removed this, since if you can right click anywhere in the file your cursor might be at a method declaration, Class name or similiar and then this would return the corresponding `PsiElement` which then tries to find the corresponding file for it, instead of the currently open file.
Is there any context where this is used @carlrobertoh? From what I can see the `IncludeFilesInContextAction` is only added to the files context-menu, for which the  `VIRTUAL_FILE_ARRAY.getData(dataContext);` works.

src/main/resources/messages/codegpt.properties
```diff
@@ -4,7 +4,7 @@ action.generateCommitMessage.title=Generate Message
 action.generateCommitMessage.description=Generate commit message
 action.generateCommitMessage.serviceWarning=Messages can only be generated with OpenAI or Azure service
 action.generateCommitMessage.missingCredentials=Credentials not provided
-action.includeFilesInContext.title=Include In Context...
+action.includeFilesInContext.title=Include File In Context...
```
PhilKes, line 7: I added 'File' here to make it clear to the user if he right clicks in the Editor it adds the File, not the text at the cursor or similar.

carlrobertoh: Perhaps it's better to have two separate keys, because otherwise the label will be wrong when you want to select a folder of files.

src/main/java/ee/carlrobert/codegpt/actions/IncludeFilesInContextAction.java
carlrobertoh: > Is there any context where this is used @carlrobertoh? From what I can see the IncludeFilesInContextAction is only added to the files context-menu, for which the VIRTUAL_FILE_ARRAY.getData(dataContext); works.

Good question. I think that for older IDE builds, the `VIRTUAL_FILE_ARRAY` didn't exist when selecting a single file, so I had to have two separate implementations. However, I believe this is no longer the case since I bumped the 'since' version a while ago.

carlrobertoh:
Thank you for adding it!

Issue #390, opened 2024-03-01
Commits genneration does not work when Custom OpenAI Service is selected


XpycTee:
### What happened?

When the Custom OpenAI service is selected, the button to generate a commit is not available with a notification:
Message can only be generated with OpenAI or Azure service.

![codegpt-commit](https://github.com/carlrobertoh/CodeGPT/assets/28292993/ab99fba5-125b-4463-b830-3d6007bc5dee)

### CodeGPT version

2.4.0

### Operating System

Windows

linpan:
me too, 
![CleanShot 2024-03-02 at 15 17 09](https://github.com/carlrobertoh/CodeGPT/assets/6077601/470f09d3-672e-4bbe-915d-3fabad724b9b)

linpan:
I will down to 2.3 version

Bikkies:
Yep, same here. It used to work fine with custom openai service
![image](https://github.com/carlrobertoh/CodeGPT/assets/29473797/5582c94f-fefd-47f0-a4f0-cebbec18c2fd)

razisam:
i have another issue with "AI commit messages". it just says something totally random, which has nothing to do with my changes:

![image](https://github.com/carlrobertoh/CodeGPT/assets/59828907/bb6f2fb8-98ba-434a-8524-e47ed8f2b07b)

XpycTee:
Hi again, updated the plugin to 2.6.3 the button became available? But now when you click on it nothing happens

Here's what comes up on the first click:

<details>
<summary>stacktreace.txt</summary>
java.lang.NoClassDefFoundError: git4idea/GitUtil
	at ee.carlrobert.codegpt.settings.configuration.BranchNamePlaceholderStrategy.getReplacementValue(Placeholder.kt:29)
	at ee.carlrobert.codegpt.settings.configuration.CommitMessageTemplate.getSystemPrompt(CommitMessageTemplate.kt:38)
	at ee.carlrobert.codegpt.actions.GenerateGitCommitMessageAction.actionPerformed(GenerateGitCommitMessageAction.java:99)
	at com.intellij.openapi.actionSystem.impl.ActionButton.actionPerformed(ActionButton.java:234)
	at com.intellij.openapi.actionSystem.impl.ActionButton.lambda$performAction$2(ActionButton.java:213)
	at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.performWithActionCallbacks(ActionManagerImpl.kt:1161)
[...]

Issue #441, opened 2024-04-09
Code completion for "Custom OpenAI Service"


boswelja:
### Describe the need of your request

I am hosting a model on my machine via Ollama, but it seems like a lot of features don't work/don't exist for "Custom OpenAI Service"? Specifically, I'd like to see code completion as an option for this service type.

### Additional context

I'm open to contributing, but it'd be nice to know about any blockers or potential issues if applicable.

reneleonhardt:
@boswelja You're right, the README isn't correct anymore since 2.6.0:
https://github.com/carlrobertoh/CodeGPT#code-completions
state `Currently supported only on GPT-3.5 and locally-hosted models`
but https://github.com/carlrobertoh/CodeGPT/commit/f0172722c75ae50d2ea895f68cbef0c90bbbcc7f#diff-dd1eb3c6e1b3ba42852ce667a754aeb087bd3d18af74e79ba8fec072bea47793R59
changed it to **OpenAI** (first) and **LLaMA C/C++** (last) service only (if their respective completions setting is enabled).

So I would suggest to mimic/duplicate those LLaMA changes to the Custom service:
* Add codeCompletionsEnabled and codeCompletionMaxTokens to CustomServiceSettingsState
* Add codeCompletionConfigurationForm to CustomServiceForm
* Add CUSTOM_OPENAI to CodeCompletionFeatureToggleActions
* Add buildCustomServiceRequest() to CodeCompletionRequestFactory
* Add CUSTOM_OPENAI to CodeGPTInlineCompletionProvider.isEnabled()
* Add useCustomOpenAIService() to ShortcutsTestMixin
* Duplicate testFetchCodeCompletionLlama() in CodeCompletionServiceTest with useCustomOpenAIService()

boswelja:
@reneleonhardt  Thanks for the detailed response! I've been poking around the codebase and realized we don't actually have a client (or API endpoint) for non-chat completions for Custom OpenAI models. There's a couple of options I can see to get around this
- Add a generic configuration for completions endpoint (same as we have now for chat completions aka "URL")
- Start moving these custom services into their own higher-level services. It's not uncommon for these services to have their own API (or an OpenAI-like API) for these, so we could save users from having to enter URLs manually
Both of these would prompt changes in https://github.com/carlrobertoh/llm-client, which I'll try to contribute regardless.
Please let me know which one you'd prefer. Personally, I'm leaning towards the second option.

reneleonhardt:
[...]

PR #476, opened 2024-04-18
Code completion for "Custom OpenAI Service"

sisve:
By popular demand; fixes(?) #441


sisve:
This should also fix #390, where commit message generation says "Messages can only be generated with OpenAI or Azure service". I've just pushed code changes that changes that tooltip text and the readme file.

src/main/kotlin/ee/carlrobert/codegpt/settings/service/custom/CustomServiceSettings.kt
```diff
@@ -0,0 +1,48 @@
+package ee.carlrobert.codegpt.settings.service.custom
+
+import com.intellij.openapi.components.*
+import com.intellij.util.xmlb.annotations.OptionTag
+import ee.carlrobert.codegpt.codecompletions.InfillPromptTemplate
+import ee.carlrobert.codegpt.util.MapConverter
+
+@Service
+@State(
+    name = "CodeGPT_CustomServiceSettings_V2",
```
carlrobertoh, line 10: This is temporary until I make it backwards-compatible

carlrobertoh:
OpenAI cannot be used at the moment because we don't support placeholders for prefix and suffix. Other than that, it seems to be working fine. The cool thing about this is that you can use two different models, I was using Llama 3 for chatting and CodeLlama 70B for code completions.

Issue #478, opened 2024-04-18
support llama3 


michabbb:
### Additional context

https://llama.meta.com/llama3/
https://github.com/meta-llama/llama3/blob/main/MODEL_CARD.md

romanstingler:
it is there I can download the model but always times out

PR #479, opened 2024-04-19
Draft: feat: Support Llama 3 model (#478)

reneleonhardt:
Wow, fresh of the press, seems to be one of the best models now!! 🚀

Waiting for llama.cpp support... 😅
https://github.com/ggerganov/llama.cpp/issues/6747
https://github.com/ggerganov/llama.cpp/pull/6751

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
+++ b/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
@@ -43,16 +43,37 @@ public enum HuggingFaceModel {
   WIZARD_CODER_PYTHON_13B_Q5(13, 5, "WizardCoder-Python-13B-V1.0-GGUF"),
   WIZARD_CODER_PYTHON_34B_Q3(34, 3, "WizardCoder-Python-34B-V1.0-GGUF"),
   WIZARD_CODER_PYTHON_34B_Q4(34, 4, "WizardCoder-Python-34B-V1.0-GGUF"),
-  WIZARD_CODER_PYTHON_34B_Q5(34, 5, "WizardCoder-Python-34B-V1.0-GGUF");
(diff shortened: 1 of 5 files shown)


carlrobertoh:
Is there any more work to be expected here, or can we perhaps remove the "Draft" prefix?

reneleonhardt:
As far as I can see it's finished (can you double-check the InfillPrompt?), I'm just waiting for the llama.cpp server support 😅
Or can you check it locally? In LM Studio I did see no big problems, the model is amazing, I just have to ask it to tell the truth when it fantasized to me about non-existing Kotlin operators 😉
7B Q8 is super fast on M1 Max 🚀

carlrobertoh:
Nice! It looks like the model doesn't support infilling, or at least I couldn't find anything. Maybe we can just remove the new infill template for now.

reneleonhardt:
> Nice! It looks like the model doesn't support infilling, or at least I couldn't find anything. Maybe we can just remove the new infill template for now.

Really? All main calls I saw contained --in-prefix and suffix https://github.com/ggerganov/llama.cpp/issues/6747#issuecomment-2065013606 that's why I added it 😅
But if you think it's a problem I can remove it for now 🙂

carlrobertoh:
Hmm, their model card doesn't specify anything related to that matter - https://llama.meta.com/docs/model-cards-and-prompt-formats/meta-llama-3

reneleonhardt:
All I can see is that there is a lot of confusion about the model, some people have many problems, some have none... I guess it dependes how you use it.
phymbert mentioned a special `infill` endpoint https://github.com/ggerganov/llama.cpp/issues/6747#issuecomment-2065117938, is CodeGPT using that?

carlrobertoh:
No, but we were using it for a while, I think. However, I can't remember the reason why we switched back.

PR #481, opened 2024-04-20
feat: add input field for llama server build parameters

### LINKED ISSUE #405: Enable BLAS/CUDA support in llama.cpp in the clion plugin

The llama.cpp automatic build that CodeGPT provides when enabling a llama model in CLion produces a server executable with no BLAS support, and unable to be activated by parameters passed to the server.

### Proposed solution

Instead of running "make" to compile llama.cpp without any of the BLAS options ideally the UI should pop up options for which BLAS library support you want to include during the compile. This is an order of magnitude speedup making operations that take minutes and time out the UI down to seconds.

LLAMA_OPENBLAS=1 for CPU only
LLAMA_CUBLAS=1 for CUDA/NVidia
LLAMA_HIPBLAS=1 for HIP/AMD
LLAMA_CLBLAST=1 for OpenCL

Alternatively there could be an option to point at the directory of an existing llama.cpp build directory

### Additional context

This can be worked around currently by browsing to the server folder for the plugin (in my case ~/.local/share/JetBrains/CLion2023.3/CodeGPT/llama.cpp) and running

```
make clean
make LLAMA_CUBLAS=1
```

Additional parameters also needs to have --n-gpu-layers set (I use a value of 49 to push the entire Code LLaMA model into vram), this would more appropriately be a specific setting defaulted to a higher value as 99% of people running the model locally will want this on.

PhilKes:
This PR closes #405 

- Adds new input field for Llama-CPP Server build parameters to e.g. enable using GPU via CUDA or similar:
image
- Improved error handling when building or running server fails:
image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -148,6 +148,7 @@ tasks {
   runIde {
     enabled = true
     environment("ENVIRONMENT", "LOCAL")
+    autoReloadPlugins.set(false) // is triggered when building llama server
   }
 
   test {
--- a/src/main/java/ee/carlrobert/codegpt/completions/llama/LlamaServerAgent.java
+++ b/src/main/java/ee/carlrobert/codegpt/completions/llama/LlamaServerAgent.java
@@ -14,14 +14,17 @@ import com.intellij.openapi.Disposable;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.components.Service;
 import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.ui.MessageType;
 import com.intellij.openapi.util.Key;
 import ee.carlrobert.codegpt.CodeGPTBundle;
 import ee.carlrobert.codegpt.CodeGPTPlugin;
 import ee.carlrobert.codegpt.settings.service.llama.LlamaSettings;
 import ee.carlrobert.codegpt.settings.service.llama.form.ServerProgressPanel;
+import ee.carlrobert.codegpt.ui.OverlayUtil;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.Consumer;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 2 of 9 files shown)


build.gradle.kts
```diff
@@ -148,6 +148,7 @@ tasks {
   runIde {
     enabled = true
     environment("ENVIRONMENT", "LOCAL")
+    autoReloadPlugins.set(false) // is triggered when building llama server
```
PhilKes, line 151: I had problems with this default feature while developing, after the llama server is compiled Intellij would sometimes try to reload the CodeGPT plugin because files in the `build/llama.cpp` folder changed, e.g.:
`2024-04-20 10:27:00,055 [  15414]   INFO - #c.i.i.p.DynamicPluginVfsListener - Detected plugin .jar file change /home/user/IdeaProjects/git/CodeGPT/build/idea-sandbox/plugins/CodeGPT/llama.cpp/llama.log, reloading plugin`

src/main/java/ee/carlrobert/codegpt/completions/llama/LlamaServerAgent.java
```diff
@@ -32,65 +35,94 @@ public final class LlamaServerAgent implements Disposable {
...
   private @Nullable OSProcessHandler startServerProcessHandler;
+  private ServerProgressPanel activeServerProgressPanel;
+  private boolean stoppedByUser;
 
   public void startAgent(
       LlamaServerStartupParams params,
       ServerProgressPanel serverProgressPanel,
       Runnable onSuccess,
-      Runnable onServerTerminated) {
+      Consumer<ServerProgressPanel> onServerTerminated) {
```
PhilKes, line 45: Changed to a `Consumer<ServerProgressPanel>` because if you:
1. Open CodeGPT Settings and start the server
2. Close the Settings
3. Reopen the Settings
4. Stop the Server, the text in the `ServerProgressPanel` is not correctly set to "Terminated Server" because an old reference to the `ServerProgressPanel` is used, therefore I changed it to always use the correct `ServerProgressPanel`instance.

src/main/java/ee/carlrobert/codegpt/completions/llama/LlamaServerAgent.java
```diff
@@ -32,65 +35,94 @@ public final class LlamaServerAgent implements Disposable {
...
+          return;
+        }
         LOG.info(event.getText());
       }
 
       @Override
       public void processTerminated(@NotNull ProcessEvent event) {
+        int exitCode = event.getExitCode();
+        LOG.info(format("Server build exited with code %d", exitCode));
+        if (stoppedByUser) {
```
PhilKes, line 105: The exit codes when killing the process on purpose seem to differ for Linux/Windows, therefore using custom flag here to determine if the process was killed on purpose.

src/main/java/ee/carlrobert/codegpt/completions/llama/LlamaServerAgent.java
```diff
@@ -32,65 +35,94 @@ public final class LlamaServerAgent implements Disposable {
...
+          onServerTerminated.accept(activeServerProgressPanel);
+          return;
+        }
+        if (exitCode != 0) {
```
PhilKes, line 109: The `make` process also prints warnings to the `stderr`, therefore the process' exit code is used to determine if there was an error

Issue #495, opened 2024-04-22
Whole line completion not working


amacfie-tc:
### What happened?

CodeGPT seems unable to show completions at the beginning of a line.  Below is a Python file in PyCharm 2024.1 (Professional Edition). When the cursor is at the very beginning of the line or after four spaces there's no completion given:
![2024-04-22-135302_458x88_scrot](https://github.com/carlrobertoh/CodeGPT/assets/91503417/61398734-ff5a-4300-b02a-d385b30bf699)

But if there are 1, 2, or 3 spaces before the cursor there is a completion:
![2024-04-22-135435_463x89_scrot](https://github.com/carlrobertoh/CodeGPT/assets/91503417/a09e5b66-a2c8-47ae-93f8-8a3834a6f029)

### CodeGPT version

2.6.3-241

### Operating System

None

carlrobertoh:
This is most likely related to how we process the response. We are simply cutting the output until the first newline is found, which obviously won't work in all situations.

PR #499, opened 2024-04-23
fix: Unsupported code completions

reneleonhardt:
Most Llama models don't support code completions (only Code Llama and Deepseek Coder).
This PR fixes the NPE that was produced and adds support for Llama 3.

I guess #495 beginning of empty line is not possible with com.intellij.codeInsight.inline.completion.elements.InlineCompletionGrayTextElement.Presentable#render returning on empty text.
But entering tabulators shows no completion either.
Any number of spaces shows completions.
But when followed by a backspace (some spaces deleted), completions disappear again 😅

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/settings/service/llama/LlamaSettings.java
+++ b/src/main/java/ee/carlrobert/codegpt/settings/service/llama/LlamaSettings.java
@@ -6,6 +6,7 @@ import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.components.PersistentStateComponent;
 import com.intellij.openapi.components.State;
 import com.intellij.openapi.components.Storage;
+import ee.carlrobert.codegpt.completions.llama.LlamaModel;
 import ee.carlrobert.codegpt.credentials.CredentialsStore;
 import ee.carlrobert.codegpt.settings.service.llama.form.LlamaSettingsForm;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


carlrobertoh:
Hmm, if infilling is not supported for Llama 3, then why support it? Can we just disable it?

reneleonhardt:
Are you sure? 😉 I tried this PR locally and code completions were shown 😅
Please try if it works for you too.

carlrobertoh:
llama3_autocomplete

carlrobertoh:
Code completions require a different prompt template

reneleonhardt:
I added a new InfillPromptTemplate, can you fix it?
Otherwise I'll remove it and just leave the NPE fixes.

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt
```diff
@@ -12,6 +12,11 @@ enum class InfillPromptTemplate(val label: String, val stopTokens: List<String>?
             return "<PRE> $prefix <SUF>$suffix <MID>"
         }
     },
+    LLAMA_3("Llama 3", listOf("<|eot_id|>")) {
+        override fun buildPrompt(prefix: String, suffix: String): String {
+            return "<|start_header_id|>user<|end_header_id|>\n\n$prefix\n$suffix<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n"
```
PhilKes, line 17: That is just the general chat prompt template for LLaMa 3 but for actual code completions we should use FIM (fill in the middle), which is not supported by any base LLama 2 / 3 model.
Should probably wait for a Codellama or similar version based on Llama 3

reneleonhardt: I'll remove it then, thank you!

PR #527, opened 2024-05-03
feat: Support CodeQwen1.5-Chat model

reneleonhardt:
https://qwenlm.github.io/blog/codeqwen1.5/

> CodeQwen1.5-7B has been pretrained with around 3 trillion tokens of code-related data. It supports an extensive repertoire of 92 programming languages, and it exhibits exceptional capacity in long-context understanding and generation with the ability to process information of 64K tokens.
In terms of performance, CodeQwen1.5 demonstrates impressive capabilities in basic code generation, long-context modelling, code editing and SQL. We believe this model can significantly enhance developer productivity and streamline software development workflows within diverse technological environments.

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
+++ b/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
@@ -4,47 +4,46 @@ import static java.lang.String.format;
 
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.regex.Pattern;
 
 public enum HuggingFaceModel {
 
-  CODE_LLAMA_7B_Q3(7, 3, "codellama-7b-instruct.Q3_K_M.gguf"),
-  CODE_LLAMA_7B_Q4(7, 4, "codellama-7b-instruct.Q4_K_M.gguf"),
-  CODE_LLAMA_7B_Q5(7, 5, "codellama-7b-instruct.Q5_K_M.gguf"),
-  CODE_LLAMA_13B_Q3(13, 3, "codellama-13b-instruct.Q3_K_M.gguf"),
-  CODE_LLAMA_13B_Q4(13, 4, "codellama-13b-instruct.Q4_K_M.gguf"),
-  CODE_LLAMA_13B_Q5(13, 5, "codellama-13b-instruct.Q5_K_M.gguf"),
-  CODE_LLAMA_34B_Q3(34, 3, "codellama-34b-instruct.Q3_K_M.gguf"),
-  CODE_LLAMA_34B_Q4(34, 4, "codellama-34b-instruct.Q4_K_M.gguf"),
-  CODE_LLAMA_34B_Q5(34, 5, "codellama-34b-instruct.Q5_K_M.gguf"),
+  CODE_LLAMA_7B_Q3(7, 3, "CodeLlama-7B-Instruct-GGUF"),
+  CODE_LLAMA_7B_Q4(7, 4, "CodeLlama-7B-Instruct-GGUF"),
+  CODE_LLAMA_7B_Q5(7, 5, "CodeLlama-7B-Instruct-GGUF"),
+  CODE_LLAMA_13B_Q3(13, 3, "CodeLlama-13B-Instruct-GGUF"),
+  CODE_LLAMA_13B_Q4(13, 4, "CodeLlama-13B-Instruct-GGUF"),
+  CODE_LLAMA_13B_Q5(13, 5, "CodeLlama-13B-Instruct-GGUF"),
+  CODE_LLAMA_34B_Q3(34, 3, "CodeLlama-34B-Instruct-GGUF"),
+  CODE_LLAMA_34B_Q4(34, 4, "CodeLlama-34B-Instruct-GGUF"),
+  CODE_LLAMA_34B_Q5(34, 5, "CodeLlama-34B-Instruct-GGUF"),
 
-  CODE_BOOGA_34B_Q3(34, 3, "codebooga-34b-v0.1.Q3_K_M.gguf"),
-  CODE_BOOGA_34B_Q4(34, 4, "codebooga-34b-v0.1.Q4_K_M.gguf"),
-  CODE_BOOGA_34B_Q5(34, 5, "codebooga-34b-v0.1.Q5_K_M.gguf"),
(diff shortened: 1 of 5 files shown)


src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
```diff
@@ -59,7 +59,30 @@ public enum HuggingFaceModel {
   PHI_3_3_8B_4K_Q5_K_S(4, 5, "Phi-3-mini-4k-instruct-Q5_K_S.gguf", "lmstudio-community", 2.82),
   PHI_3_3_8B_4K_Q6_K(4, 6, "Phi-3-mini-4k-instruct-Q6_K.gguf", "lmstudio-community", 3.14),
   PHI_3_3_8B_4K_Q8_0(4, 8, "Phi-3-mini-4k-instruct-Q8_0.gguf", "lmstudio-community", 4.06),
-  PHI_3_3_8B_4K_FP16(4, 16, "Phi-3-mini-4k-instruct-fp16.gguf", "lmstudio-community", 7.64);
+  PHI_3_3_8B_4K_FP16(4, 16, "Phi-3-mini-4k-instruct-fp16.gguf", "lmstudio-community", 7.64),
+
+  CODE_QWEN_1_5_7B_Q2_K(7, 2, "CodeQwen1.5-7B-Chat.Q2_K.gguf", "RichardErkhov", 3.05),
```
PhilKes, line 64: I checked with @carlrobertoh and we agreed there is no benefit in listing all available quantizations here, for https://github.com/carlrobertoh/CodeGPT/pull/529/files we agreed that it is enough to only offer  `K_M` quantizations for 3-8 Bit (if there is no `K_M`, use only `K` or `M` ones)

src/main/java/ee/carlrobert/codegpt/completions/llama/LlamaModel.java
```diff
@@ -113,7 +113,41 @@ public enum LlamaModel {
           HuggingFaceModel.PHI_3_3_8B_4K_Q5_K_S,
           HuggingFaceModel.PHI_3_3_8B_4K_Q6_K,
           HuggingFaceModel.PHI_3_3_8B_4K_Q8_0,
-          HuggingFaceModel.PHI_3_3_8B_4K_FP16));
+          HuggingFaceModel.PHI_3_3_8B_4K_FP16)),
+  CODE_QWEN(
+      "CodeQwen1.5-7B", """
```
PhilKes, line 118: The "1.5-7B" should be removed from the label, since its redundant in the UI:
image

PhilKes (commented):
CodeQwen also offers FIM capabilities, I tested it with the following `InfillPromptTemplate`:
```java
CODE_QWEN("CodeQwen", listOf("<|endoftext|>")) {
        override fun buildPrompt(prefix: String, suffix: String): String {
            return "<fim_prefix>$prefix<fim_suffix>$suffix<fim_middle>"
        }
    };
```

src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
reneleonhardt: So I should also drop Q2_K?

src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
```diff
@@ -4,47 +4,46 @@
 
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.regex.Pattern;
 
 public enum HuggingFaceModel {
 
-  CODE_LLAMA_7B_Q3(7, 3, "codellama-7b-instruct.Q3_K_M.gguf"),
```
PhilKes, line 11: I guess you made a mistake while rebasing/merging, these changes should not be reverted, they were part of the https://github.com/carlrobertoh/CodeGPT/pull/529 fix

reneleonhardt:
I fixed the broken download URLs by reverting master, please double-check by clicking each model link:
```java
public static void main(String[] args) {
  java.util.Arrays.stream(values())
    .collect(java.util.stream.Collectors.groupingBy(e -> e.name().substring(0, 10),
      java.util.stream.Collectors.toList()))
    .values().forEach(e -> System.out.println(e.get(0).getFileURL()));
}
```

src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
reneleonhardt: As I commented, the links were broken, now every URL is working again.

reneleonhardt: But if you insist, I reapply master, not my responsibility then ;)

PhilKes: The `modelName` is used to store the selected model in the XML config, so we can't have different models having the same `modelName`, e.g:
https://github.com/carlrobertoh/CodeGPT/blob/eb595a73c0a4ca81753d5248a622be8be16537c0/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java#L10-L11

We either have to manually concatenate the quantization to the `toString` value or introduce a new field for the download directory.

reneleonhardt: Done, so all URLs are working except TheBloke (new master), good night 😅

PhilKes: ´toString` could probably also just return `name()`, since the enum names are unique

reneleonhardt: Why do you merge such a breaking change without testing it, at least manually?
```java
public static void main(String[] args) {
  java.util.Arrays.stream(values())
    .collect(java.util.stream.Collectors.groupingBy(e -> e.directory, java.util.stream.Collectors.toList()))
    .values().stream().map(e -> e.get(0)).sorted()
    .forEach(e -> System.out.println(e.getFileURL()));
}

```

PhilKes: As I said I'm sorry I didn't notice the breaking change, @carlrobertoh also didn't notice it when reviewing it, it also hasnt been released yet either, so no need to make a big fuss out of it 😄

PR #529, opened 2024-05-07
fix: Storing HuggingFaceModel by modelName instead of quantization only

PhilKes:
This PR fixes how the `HuggingFaceModel` for Llama is stored.
Scenario:
Select `CodeGemma` `4-bit precision` -> Save

`CodeGPT_LlamaSettings.xml` Before:
```xml
<!-- ambiguous -->
<option name="huggingFaceModel" value="4-bit precision" /> 
```

After:
```xml
<option name="huggingFaceModel" value="codegemma-7b-it-Q4_K_M.gguf" />
```

Also removed unnecessary `CodeGemma` and `Phi-3` models


src/main/java/ee/carlrobert/codegpt/settings/service/llama/LlamaSettings.java
```diff
@@ -26,6 +28,17 @@ public LlamaSettingsState getState() {
   @Override
   public void loadState(@NotNull LlamaSettingsState state) {
     this.state = state;
+    // Catch if model's name has changed which could lead to
+    // HuggingFaceModel or PromptTemplates not being found
+    if (this.state.getHuggingFaceModel() == null) {
```
carlrobertoh, line 33: What kind of compatibility issues does it solve? I'm not quite sure I'm following.

carlrobertoh: Right! Thank you for explaining :)

PR #531, opened 2024-05-07
fix: CustomService Test connection with correct settings

### LINKED ISSUE #528: Ollama support from Intellij
### What happened?

Trying to configure Ollama from intellij. Two problems:
1. UI is looking for API key
2. Once putting API key it failed in text connection

image

### Relevant log output or stack trace

_No response_

### Steps to reproduce

Install codegpt on Intellij. Try to configure it with CustomOpenAI and Ollama. Put API key and test connection

### CodeGPT version

2.6.3-241

### Operating System

macOS

PhilKes:
This PR fixes #528 
The `Test Connection` in the `CustomServiceForm` was using the currently saved settings instead of the settings from the currently open UI, therefore leading to confusing Error messages, if you e.g. switch from Azure to Ollama and then clicking `Test Connection` without hitting `Apply` before.

### DIFF
--- a/src/main/kotlin/ee/carlrobert/codegpt/settings/service/custom/CustomServiceChatCompletionForm.kt
+++ b/src/main/kotlin/ee/carlrobert/codegpt/settings/service/custom/CustomServiceChatCompletionForm.kt
@@ -15,7 +15,10 @@ import javax.swing.JButton
 import javax.swing.JPanel
 import javax.swing.SwingUtilities
 
-class CustomServiceChatCompletionForm(state: CustomServiceChatCompletionSettingsState) {
+class CustomServiceChatCompletionForm(
+    state: CustomServiceChatCompletionSettingsState,
+    val getApiKey: () -> String?
+) {
 
     private val urlField = JBTextField(state.url, 30)
     private val tabbedPane = CustomServiceFormTabbedPane(state.headers, state.body)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/ee/carlrobert/codegpt/settings/service/custom/CustomServiceChatCompletionForm.kt
```diff
@@ -15,7 +15,10 @@ import javax.swing.JButton
 import javax.swing.JPanel
 import javax.swing.SwingUtilities
 
-class CustomServiceChatCompletionForm(state: CustomServiceChatCompletionSettingsState) {
+class CustomServiceChatCompletionForm(
+    state: CustomServiceChatCompletionSettingsState,
+    val getApiKey: () -> String?
```
carlrobertoh, line 20: How about we pass the reference of `apiKeyField` instead and then just extract the key directly from the field inside the listener?

PhilKes: We could, but then we would essentially just copy&paste the `getApiKey` into `CustomServiceChatCompletionForm`
https://github.com/carlrobertoh/CodeGPT/blob/4c2639d1bfad92ea7df3bd20079512b806071ddb/src/main/kotlin/ee/carlrobert/codegpt/settings/service/custom/CustomServiceForm.kt#L93

I didnt see any benefit to it, also in terms of capsulation I thought its cleaner if the `CustomServiceChatCompletionForm` does not have direct access to the `apiKeyField` Component.

carlrobertoh: Sure, I'm fine either way. I usually forward the callbacks only when the consumer needs to handle the component's actions/events.

PR #533, opened 2024-05-07
Revert "fix: use /infill for llama.cpp code-completions (#513)"

PhilKes:
This reverts commit 8de72b330178fa97b0482e1dbb2829964f8b737a.

As discussed in [#510 ](https://github.com/carlrobertoh/CodeGPT/pull/510#issuecomment-2096486472) this reverts the switch from `/completion` to `/infill`

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/completions/CompletionRequestService.java
+++ b/src/main/java/ee/carlrobert/codegpt/completions/CompletionRequestService.java
@@ -120,7 +120,7 @@ public final class CompletionRequestService {
           CodeCompletionRequestFactory.buildCustomRequest(requestDetails),
           new OpenAITextCompletionEventSourceListener(eventListener));
       case LLAMA_CPP -> CompletionClientProvider.getLlamaClient()
-          .getInfillAsync(
+          .getChatCompletionAsync(
               CodeCompletionRequestFactory.buildLlamaRequest(requestDetails),
               eventListener);
       default ->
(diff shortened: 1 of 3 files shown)


CISC:
Hey, just curious why [ggerganov/llama.cpp#7102 (comment)](https://github.com/ggerganov/llama.cpp/issues/7102#issuecomment-2095956124) prompted this?

PhilKes:
> Hey, just curious why [ggerganov/llama.cpp#7102 (comment)](https://github.com/ggerganov/llama.cpp/issues/7102#issuecomment-2095956124) prompted this?

We use GGUF models form HF (e.g https://huggingface.co/Qwen/CodeQwen1.5-7B-Chat-GGUF) which (afaik) do not have the special token modification from your PR (https://github.com/ggerganov/llama.cpp/pull/7166), therefore we can't use them with `/infill`, and reverted back to using `/completion` for FIM prompts.

CISC:
You can use [mine](https://huggingface.co/CISCai/CodeQwen1.5-7B-Chat-SOTA-GGUF) which does. :)

PhilKes:
Thats nice, but we would need a working GGUF for **all** models that we support, the list is quite long (see [HuggingFaceModel](https://github.com/carlrobertoh/CodeGPT/blob/f260a71d1b6e1a61731cd1649b4fec64e78064ea/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java)) and its continously growing with new models coming up. We want the same solution for every existing and new model, so `/infill` is no option for us atm

CISC:
In that case I think you will have to maintain your own copies, which shouldn't be that hard though using `gguf-new-metadata.py`. I doubt anyone else is going to bother manually adding this metadata, and it's unlikely the conversion scripts will (it's not quite working properly on the ones it does add them to already (instruct/chat tuned models can lose fill-in-middle capability even though they still have the tokens)).

PhilKes:
> In that case I think you will have to maintain your own copies, which shouldn't be that hard though using `gguf-new-metadata.py`. I doubt anyone else is going to bother manually adding this metadata, and it's unlikely the conversion scripts will (it's not quite working properly on the ones it does add them to already (instruct/chat tuned models can lose fill-in-middle capability even though they still have the tokens)).

Thanks for your suggestion but its easier for us to just maintain the [FIM templates](https://github.com/carlrobertoh/CodeGPT/blob/f260a71d1b6e1a61731cd1649b4fec64e78064ea/src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt) in our project and using the `/completion` Endpoint, than maintaining our own GGUF copies.

PR #543, opened 2024-05-11
feat: Visualize downloaded models

reneleonhardt:
With more and more LLaMA models users should have a clear path to select/switch one of their downloaded models in as few clicks as possible.
Select Quantization
Select Model


carlrobertoh:
Nice!

PR #544, opened 2024-05-12
feat: Start/stop LLaMA Server from statusbar

reneleonhardt:
### Features
* User can start or stop the server from the statusbar
* User can start or stop the server with the hotkey `ctrl + shift + alt + S` (`command + shift + alt + S` on Mac)
* Actions not available / hidden on Windows, or if another service is selected, or if the selected model isn't downloaded
* Notifications describe the state (Starting, Running, Stopping, Stopped)
* Only one notification is visible at any time even when you toggle the action again repeatedly

### Fixed / Improved
* LLaMA model selected when settings are opened
* Statusbar actions use localized messages
* Streamlined paths usage
* Refactored preferences form and notifications

### Suggestions
* Rename terminated to stopped (reused preferences wording)

Start Server
Stop Server
Starting a server
Server running
Server terminated


carlrobertoh:
I think we should make it visible only to Mac and Linux users, since we don't support running the server on Windows yet.

reneleonhardt:
> I think we should make it visible only to Mac and Linux users, since we don't support running the server on Windows yet.

Good idea, I will refactor the code tonight.

carlrobertoh:
> > I think we should make it visible only to Mac and Linux users, since we don't support running the server on Windows yet.
> 
> Good idea, done. Can I change `terminated` to `stopped` everywhere? 😅

Sure, "stopped" sounds better.

PR #552, opened 2024-05-15
feat: Support Stable Code Instruct 3B

reneleonhardt:
https://stability.ai/news/introducing-stable-code-instruct-3b
https://huggingface.co/stabilityai/stable-code-instruct-3b
https://huggingface.co/bartowski/stable-code-instruct-3b-GGUF
https://ollama.com/library/stable-code

Prompt seems identical to Code Qwen, but I had to add `<|im_end|>` as stop token.
Can you try it out, FIM isn't possible with the instruct model, right?

Too bad we can't run 2 models after each other like I proposed...
* User enters chat window focus, start Instruct model automatically (stop FIM if running)
* User leaves chat window focus, start Code Completion model automatically (stop Instruct if running)

For example
https://huggingface.co/google/codegemma-7b
https://huggingface.co/google/codegemma-7b-it
https://huggingface.co/meta-llama/Meta-Llama-3-8B
https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
+++ b/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
@@ -1,5 +1,6 @@
 package ee.carlrobert.codegpt.completions;
 
+import static ee.carlrobert.codegpt.completions.HuggingFaceModel.Model.SC3;
 import static ee.carlrobert.codegpt.completions.llama.LlamaModel.getDownloadedMarker;
 import static ee.carlrobert.codegpt.completions.llama.LlamaModel.getLlamaModelsPath;
 import static java.lang.String.format;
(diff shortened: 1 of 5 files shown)


carlrobertoh:
FIM seems to be possible with this model. However, you need to specify the correct FIM template in order for this to work.

> It supports code completion, FIM tasks, and understands nuanced programming queries, enabling developers to work more efficiently and effectively with less effort. 

https://stability.ai/news/introducing-stable-code-instruct-3b

Screenshot 2024-05-15 at 10 45 35

reneleonhardt:
I reused the CodeQwen infill template (like the prompt template), it seems to work.

In addition I sorted the LLaMA models in settings because it seems fairer than appending all new models to the end where no one can see them... it would be cool to persist the plugin version between updates/reinstall, then we could mark all new ones (according to each user personally).
Similar to the changelog, now IntelliJ showed only What's New 2.7.1, but I was coming from 2.6.3, so I can't see what changed in 2.7.0... strange "changelog" only showing one change/release 🤷

carlrobertoh:
> it would be cool to persist the plugin version between updates/reinstall, then we could mark all new ones (according to each user personally).

Not sure I follow. What do you mean by persisting the plugin version between updates?

> Similar to the changelog, now IntelliJ showed only What's New 2.7.1, but I was coming from 2.6.3, so I can't see what changed in 2.7.0... strange "changelog" only showing one change/release 🤷

Yeah, it probably just shows the latest/recent changes. I had to patch a fix really quickly due to some backward compatibility issues.

PR #571, opened 2024-05-17
feat: add project context to code completions

PhilKes:
**WORK IN PROGRESS**

This PR adds context to the code-completions.

Quite a few of the models support repository-level completions via special tokens:
[CodeQwen 1.5](https://github.com/QwenLM/CodeQwen1.5?tab=readme-ov-file#2-file-level-code-completion-fill-in-the-middle)
[DeepSeek Coder](https://github.com/deepseek-ai/DeepSeek-Coder?tab=readme-ov-file#2-code-insertion)
[CodeGemma](https://huggingface.co/google/codegemma-7b#for-code-completion)
[StarCoder2](https://huggingface.co/spaces/bigcode/bigcode-playground/blob/main/app.py)

Consider this simple Java example from an old project of mine ([movieland](https://github.com/PhilKes/movieland)).
`MovieUtils.java`:
```java
@Service
public class MovieUtils {
    private final MovieService movieService;

    public MovieUtils(MovieService movieService) {
        this.movieService = movieService;
    }
    public List<Movie> getMoviesSortedByPlaytime(String name) {
        return <CURSOR>
    }
     
}
```
Obviously in examples like this, the code-completion without context can not be of much help since the model doesnt know the contents of the [MovieService](https://github.com/PhilKes/movieland/blob/master/movieland-backend/src/main/java/com/phil/movieland/rest/service/MovieService.java) or [Movie](https://github.com/PhilKes/movieland/blob/master/movieland-backend/src/main/java/com/phil/movieland/data/entity/Movie.java) class of my project.

## Proposal 
To get further context for a specific code-completion example we have to somehow parse the given code and then decide what other project files could be useful.
To do that, I am leveraging the **IntelliJ OpenAPI's Program Structure Interface ([PSI](https://plugins.jetbrains.com/docs/intellij/psi.html))**, which the IDEs use internally to parse the source code files.
IntelliJ provides a `PSI` implementation for every programming language they support (see [IntelliJ Modules](https://plugins.jetbrains.com/docs/intellij/plugin-compatibility.html#modules-specific-to-functionality))
These are the steps to determine the context files for a java project:
1. Determine the `PsiElement` at the cursor position (in the above example: `PsiWhiteSpace`)
2. Find the closest parent `PsiMethod` or `PsiClass` (in the above example: `PsiMethod:getMoviesSortedByPlaytime`)
3. Determine all `PsiTypeElement`s that are used inside of that parent (e.g. method parameters, return value type, local variables) (in the above example: `PsiTypeElement:Movie`, `PsiTypeElement:String`)
4. If the parent is a `PsiMethod` also class and static fields of the parent `PsiClass` are taken into account  (see example `movieService` field) (in the above example: `PsiTypeElement:MovieService`)
5. For all found `PsiTypeElement`s the `VirtualFile` containing its source code is evaluated (only for project files, library or jdk files/classes are ignored) (in the above example: `Movie.java`, `MovieService.java`)
6. The contents of the found `VirtualFile`s are prepended to the code-completion prompts according to the models' prompt formats for repository-level completions.

## Results
I have tested the described example with all models that support repository-level completions and tried to accumulate the results here: [Code_Completions_Context_Model Google Sheet](https://docs.google.com/spreadsheets/d/e/2PACX-1vTpuif13yyXrEK9M3XVTjwHV8DDSZ3qmolfGoUGSA_6pk1kSEfUQSoGP2UcsKrIFD8XGmb3TEX7BPaB/pubhtml). Be aware that I only tested the described scenario and did not do large test runs, the measurement were all done once to get a basic overview of the performance.

_Note: I am not sure yet why the DeepSeekCoder 6.7b tests without context returned absolute jibberish_

For all models I tested the code-completions as they were and then with the new context included.
As you can see in the spread sheet, the **completions without context gave terrible results** with method calls on the `movieService` instance that do not even exist.
The **best result was achieved with `CodeQwen 1.5 (IQ4_NL)` with context**, which returned a great, working completion:
```java
public List<Movie> getMoviesSortedByPlaytime(String name) {
    return movieService.queryAllMovies(name).stream() .sorted(Comparator.comparingLong(Movie::getLength
}
```
It found the correct method `MovieService.queryAllMovies` to use and also understood the `Movie.length` field is interpreted as its playtime.

_Note: I disabled the code-completion from being stopped at "\n" for this test, but didn't increase the `maxTokens`_

### Advantages:
- Much better completions in many scenarios
- Some smaller models (e.g. DeepSeekCoder 1.3b-instruct.Q5_K_M) returned quite good results while still having a low response time (~2s)
- Newer models such as `CodeQwen 1.5` performs very good with context size up to 64k tokens (see https://qwenlm.github.io/blog/codeqwen1.5/#codeqwen-are-long-context-coders)
- In theory we could also add library source code to the prompt since they can be evaluated with `PSI` as well. Since models are only trained on specific timestamps, this could be used to "keep them up-to-date" with newer library source code

### Disadvantages:
- Prompts have much more tokens (e.g. 131 without context vs 2024 with context), therefore also higher response times
- Implementation to retrieve context is language specific (e.g. have to provide different implementation for Java, Python, JavaScript)

## Open Questions
- Which model is the best compromise between response time and completion quality?
- Which level of depth should the context include?
- Implement the context search for other languages than Java
- Should the user have an option to include/exclude this automatic context for code-completions?


src/main/resources/META-INF/plugin.xml
```diff
@@ -5,6 +5,13 @@
     <depends>com.intellij.modules.platform</depends>
     <depends>com.intellij.modules.lang</depends>
     <depends optional="true" config-file="plugin-java.xml">com.intellij.modules.java</depends>
+    <depends optional="true" config-file="plugin-python.xml">com.intellij.modules.python</depends>
+<!--    TODO-->
+<!--    <depends optional="true" config-file="plugin-js.xml">JavaScript</depends>-->
```
PhilKes, line 10: These would be needed to implement a `LanguagePsiParser` for e.g. JavaScript, PHP, ...
In the `plugin-js.xml` a `JavaScriptPsiParser` service would be added, which would only be loaded if the user uses IntelliJ Ultimate or WebStorm (which have JS support)

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/psi/PythonContextFinder.kt
```diff
@@ -0,0 +1,76 @@
...
+     * Finds enclosing [PyFunction] or [PyClass] of [psiElement] and
+     * determines source code files of all used [PyReferenceExpression]s for the context.
+     */
+    override fun findContextSourceFiles(psiElement: PsiElement, editor: Editor): Set<VirtualFile> {
+        var significantElement = psiElement
+        // In Python whitespaces have semantic meaning
+        // its enclosing PyFunction is not inevitably its parent node, but its previous sibling
+        while (significantElement is PsiWhiteSpace) {
+            significantElement = significantElement.prevSibling
+        }
```
PhilKes, line 26: There is probably room for refactoring/optimization here, to e.g. use [PsiTreeUtil.getPrevSiblingOfType](https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/psi/util/PsiTreeUtil.java#L511)

PhilKes:
Another great thing about PSI, we can not only use IntelliJ's language support, IntellIJ Ultimate also offers e.g. a [Spring API](https://plugins.jetbrains.com/docs/intellij/spring-api.html#beans). This offers the ability to access a `CommonSpringModel` which allows us to easily find e.g. Spring Beans (`SpringModelSearchers#findBeans`) that are used to provide further context for the completion.

PhilKes:
image
The JetBrains IDEs also offer a neat little tool to view the `PSI` tree of a file (`Tools` -> `View PSI Structure of Current File`, which could be very helpful in extracting the correct `PsiElement`s

carlrobertoh:
This is really awesome work and one of the missing pieces of CodeGPT! 🚀 💯 

Few comments:

> Note: I disabled the code-completion from being stopped at "\n" for this test, but didn't increase the maxTokens

Hmm, cancelling the request at '\n' was removed in the latest release. Also, `maxTokens` are calculated dynamically instead of letting the user decide. The current solution is somewhat funky, but it works quite well in some situations.

> In theory we could also add library source code to the prompt since they can be evaluated with PSI as well. Since models are only trained on specific timestamps, this could be used to "keep them up-to-date" with newer library source code

Yes, and I think we should already include this change in the current branch. This shouldn't require any additional implementation IF you remove the `PsiTarget` -> `VirtualFile` mapping and take the content directly from the PSI element.

> Prompts have much more tokens (e.g. 131 without context vs 2024 with context), therefore also higher response times

This is inevitable, we just need to find the perfect balance.

> Which model is the best compromise between response time and completion quality?

Currently, most users are using the OpenAI provider, which means that `gpt-3.5-turbo-instruct` should be considered the primary model for code completions.

> Should the user have an option to include/exclude this automatic context for code-completions?

I think so, similarily to how we allow users to enable/disable completion post-processing. Eventually, we can remove them altogether, as they can be relatively hard for users to understand.

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt
```diff
@@ -1,42 +1,97 @@
...
+    },
+    STAR_CODER("StarCoder2", listOf("<|endoftext|>")) {
+        override fun buildPrompt(infillDetails: InfillRequestDetails): String {
+            // see https://huggingface.co/spaces/bigcode/bigcode-playground/blob/main/app.py
+            val infillPrompt =
+                "<fim_prefix>${infillDetails.prefix} <fim_suffix> ${infillDetails.suffix}<fim_middle>"
+            return if (infillDetails.context == null || infillDetails.context.contextFiles.isNullOrEmpty()) {
+                infillPrompt
+            } else {
+                "<reponame>${infillDetails.context.repoName}" +
```
carlrobertoh, line 83: This is really cool! I didn't know it was possible to give in additional knowledge like this 👍

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillRequestDetails.kt
```diff
@@ -1,55 +1,53 @@
...
             caretOffset: Int,
-            start: Int,
-            end: Int,
+            prefix: String,
+            suffix: String
         ): InfillRequestDetails {
-            val prefix: String = truncateText(document, start, caretOffset, false)
-            val suffix: String = truncateText(document, caretOffset, end, true)
-            return InfillRequestDetails(prefix, suffix)
+            val truncatedPrefix = prefix.takeLast(MAX_OFFSET)
```
carlrobertoh, line 25: We need to find a better solution for this `MAX_OFFSET` because 10,000 chars is a lot, especially since we're now providing more additional context.

PhilKes:
> > In theory we could also add library source code to the prompt since they can be evaluated with PSI as well. Since models are only trained on specific timestamps, this could be used to "keep them up-to-date" with newer library source code
> 
> Yes, and I think we should already include this change in the current branch. This shouldn't require any additional implementation IF you remove the `PsiTarget` -> `VirtualFile` mapping and take the content directly from the PSI element.
> >
Getting the text contents from the `PsiTarget` does work, but for library code it will not return the source code, but the compiled code. For example if I have a reference to Jackson's `ObjectMapper` the text content of its `PsiTarget`/`PsiClassImpl` is:


<details>
  <summary>PsiClassImpl#ObjectMapper</summary>
  
```java
public class ObjectMapper extends com.fasterxml.jackson.core.ObjectCodec implements com.fasterxml.jackson.core.Versioned, java.io.Serializable {
    private static final long serialVersionUID = 2L;
    protected static final com.fasterxml.jackson.databind.AnnotationIntrospector DEFAULT_ANNOTATION_INTROSPECTOR;
    protected static final com.fasterxml.jackson.databind.cfg.BaseSettings DEFAULT_BASE;
    protected final com.fasterxml.jackson.core.JsonFactory _jsonFactory;
    protected com.fasterxml.jackson.databind.type.TypeFactory _typeFactory;
    protected com.fasterxml.jackson.databind.InjectableValues _injectableValues;
    protected com.fasterxml.jackson.databind.jsontype.SubtypeResolver _subtypeResolver;
    protected final com.fasterxml.jackson.databind.cfg.ConfigOverrides _configOverrides;
    protected final com.fasterxml.jackson.databind.cfg.CoercionConfigs _coercionConfigs;
    protected com.fasterxml.jackson.databind.introspect.SimpleMixInResolver _mixIns;
    protected com.fasterxml.jackson.databind.SerializationConfig _serializationConfig;
    protected com.fasterxml.jackson.databind.ser.DefaultSerializerProvider _serializerProvider;
    protected com.fasterxml.jackson.databind.ser.SerializerFactory _serializerFactory;
    protected com.fasterxml.jackson.databind.DeserializationConfig _deserializationConfig;
    protected com.fasterxml.jackson.databind.deser.DefaultDeserializationContext _deserializationContext;
    protected java.util.Set<java.lang.Object> _registeredModuleTypes;
    protected final java.util.concurrent.ConcurrentHashMap<com.fasterxml.jackson.databind.JavaType,com.fasterxml.jackson.databind.JsonDeserializer<java.lang.Object>> _rootDeserializers;

    public ObjectMapper() { /* compiled code */ }

    public ObjectMapper(com.fasterxml.jackson.core.JsonFactory jf) { /* compiled code */ }

    protected ObjectMapper(com.fasterxml.jackson.databind.ObjectMapper src) { /* compiled code */ }

    protected ObjectMapper(com.fasterxml.jackson.databind.ObjectMapper src, com.fasterxml.jackson.core.JsonFactory factory) { /* compiled code */ }

    public ObjectMapper(com.fasterxml.jackson.core.JsonFactory jf, com.fasterxml.jackson.databind.ser.DefaultSerializerProvider sp, com.fasterxml.jackson.databind.deser.DefaultDeserializationContext dc) { /* compiled code */ }

    protected com.fasterxml.jackson.databind.introspect.ClassIntrospector defaultClassIntrospector() { /* compiled code */ }

    public com.fasterxml.jackson.databind.ObjectMapper copy() { /* compiled code */ }

    public com.fasterxml.jackson.databind.ObjectMapper copyWith(com.fasterxml.jackson.core.JsonFactory factory) { /* compiled code */ }

    protected void _checkInvalidCopy(java.lang.Class<?> exp) { /* compiled code */ }

    protected com.fasterxml.jackson.databind.ObjectReader _newReader(com.fasterxml.jackson.databind.DeserializationConfig config) { /* compiled code */ }

    protected com.fasterxml.jackson.databind.ObjectReader _newReader(com.fasterxml.jackson.databind.DeserializationConfig config, com.fasterxml.jackson.databind.JavaType valueType, java.lang.Object valueToUpdate, com.fasterxml.jackson.core.FormatSchema schema, com.fasterxml.jackson.databind.InjectableValues injectableValues) { /* compiled code */ }

    protected com.fasterxml.jackson.databind.ObjectWriter _newWriter(com.fasterxml.jackson.databind.SerializationConfig config) { /* compiled code */ }

    protected com.fasterxml.jackson.databind.ObjectWriter _newWriter(com.fasterxml.jackson.databind.SerializationConfig config, com.fasterxml.jackson.core.FormatSchema schema) { /* compiled code */ }

    protected com.fasterxml.jackson.databind.ObjectWriter _newWriter(com.fasterxml.jackson.databind.SerializationConfig config, com.fasterxml.jackson.databind.JavaType rootType, com.fasterxml.jackson.core.PrettyPrinter pp) { /* compiled code */ }

    public com.fasterxml.jackson.core.Version version() { /* compiled code */ }

    public com.fasterxml.jackson.databind.ObjectMapper registerModule(com.fasterxml.jackson.databind.Module module) { /* compiled code */ }
...
}
```
</details>

I think to get the actual source code it has to be downloaded first (for e.g. java usually libraries provide separate `sources.jar`, which is downloaded by [DownloadSourcesAction](https://github.com/JetBrains/intellij-community/blob/57f19ccb362b9dedc4fb374c58cf2bfd756a3b3a/java/idea-ui/src/com/intellij/jarFinder/AbstractAttachSourceProvider.java#L111) 


> > Which model is the best compromise between response time and completion quality?
> 
> Currently, most users are using the OpenAI provider, which means that `gpt-3.5-turbo-instruct` should be considered the primary model for code completions.
> 
> >
I haven't tested this with ChatGPT, as far as I know `gpt-3.5-turbo-instruct` was only trained with the FIM tokens, not repository-level completion tokens like the mentioned open source models. We would have to come up with our own syntax for prompting ChatGPT, although I am not sure how reliable that would be. 

> > Should the user have an option to include/exclude this automatic context for code-completions?
> 
> I think so, similarly to how we allow users to enable/disable completion post-processing. Eventually, we can remove them altogether, as they can be relatively hard for users to understand.
> >
I agree, I think customizability has been mentioned many times as one of the advantages of the CodeGPT plugin, so we should keep doing that. It will also take some time to find the best configuration either way.

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt
```diff
@@ -1,42 +1,97 @@
...
+        override fun buildPrompt(infillDetails: InfillRequestDetails): String {
+            // see https://huggingface.co/google/codegemma-7b#for-code-completion
+            val infillPrompt =
+                "<|fim_prefix|>${infillDetails.prefix}<|fim_suffix|>${infillDetails.suffix}<|fim_middle|>"
+            return if (infillDetails.context == null || infillDetails.context.contextFiles.isNullOrEmpty()) {
+                infillPrompt
+            } else {
+                infillDetails.context.contextFiles.map {
+                    "<|file_separator|>${it.path} \n" +
+                            it.readText()
```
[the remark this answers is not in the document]
PhilKes: Do you mean to if e.g. there is a method call to [StringUtils.capitalize](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#capitalize-java.lang.String-) that only the source code of the method should be included, not its entire class?
Yes that would be possible, although I am not sure if the models will get confused by that, because from what I read during training they always provided the **entire** source code file, not just "relevant" parts (for the models with repository-level completion tokens). But I would also prefer just to send the relevant parts obviously.

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillRequestDetails.kt
PhilKes: Yes definetely, perhaps we need to leverage the PSI tree to limit the completion to e.g. method or class level, depending on how big the file is. I remember I implemented the code completions in the beginning with limiting it to method-level instead of sending the entire file

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt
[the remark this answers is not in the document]
PhilKes: There is a method [ClsMethodImpl.sourceMirrorMethod](https://github.com/JetBrains/intellij-community/blob/57f19ccb362b9dedc4fb374c58cf2bfd756a3b3a/java/java-psi-impl/src/com/intellij/psi/impl/compiled/ClsMethodImpl.java#L247) which seems to return the source code of the method as a `PsiElement`. If you did not yet use the  "Download sources" action it returns `null`, but if you did it return this, e.g. for [StringUtils.capitalize](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#capitalize-java.lang.String-):

<details>
  <summary>ClsMethodImpl.sourceMirrorMethod.text</summary>
  
  ```java
/**
     * <p>Capitalizes a String changing the first character to title case as
     * per {@link Character#toTitleCase(int)}. No other characters are changed.</p>
     *
     * <p>For a word based algorithm, see {@link org.apache.commons.lang3.text.WordUtils#capitalize(String)}.
     * A {@code null} input String returns {@code null}.</p>
     *
     * <pre>
     * StringUtils.capitalize(null)  = null
     * StringUtils.capitalize("")    = ""
     * StringUtils.capitalize("cat") = "Cat"
     * StringUtils.capitalize("cAt") = "CAt"
     * StringUtils.capitalize("'cat'") = "'cat'"
     * </pre>
     *
     * @param str the String to capitalize, may be null
     * @return the capitalized String, {@code null} if null String input
     * @see org.apache.commons.lang3.text.WordUtils#capitalize(String)
     * @see #uncapitalize(String)
     * @since 2.0
     */
    public static String capitalize(final String str) {
        final int strLen = length(str);
        if (strLen == 0) {
            return str;
        }

        final int firstCodepoint = str.codePointAt(0);
        final int newCodePoint = Character.toTitleCase(firstCodepoint);
        if (firstCodepoint == newCodePoint) {
            // already capitalized
            return str;
        }

        final int[] newCodePoints = new int[strLen]; // cannot be longer than the char array
        int outOffset = 0;
        newCodePoints[outOffset++] = newCodePoint; // copy the first codepoint
        for (int inOffset = Character.charCount(firstCodepoint); inOffset < strLen; ) {
            final int codepoint = str.codePointAt(inOffset);
            newCodePoints[outOffset++] = codepoint; // copy the remaining ones
            inOffset += Character.charCount(codepoint);
         }
        return new String(newCodePoints, 0, outOffset);
    }
```
</details>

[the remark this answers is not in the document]
Comment (carlrobertoh) [reply src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt:30]: > Do you mean to if e.g. there is a method call to StringUtils.capitalize that only the source code of the method should be included, not its entire class?

Exactly!

> Yes that would be possible, although I am not sure if the models will get confused by that, because from what I read during training they always provided the entire source code file, not just "relevant" parts (for the models with repository-level completion tokens). But I would also prefer just to send the relevant parts obviously.

I tried sending snippets of code before the prefix of a cursor, and `gpt-3.5-instruct` didn't seem to struggle with it, given that the entire context was syntactically incorrect. However, it may be different per model.

File: src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillRequestDetails.kt (hunk 4 above)
Comment (carlrobertoh) [reply src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillRequestDetails.kt:25]: Yeah, I remember something similar. However, having just a method-level context isn't sufficient without the relevant context.

**Comment by carlrobertoh** [comment]
> Getting the text contents from the PsiTarget does work, but for library code it will not return the source code, but the compiled code. For example if I have a reference to Jackson's ObjectMapper the text content of its PsiTarget/PsiClassImpl is:

Ahh, ok. In that case, can we include the functionality only if the user has downloaded the source code?

**Comment by PhilKes** [comment]
> > Getting the text contents from the PsiTarget does work, but for library code it will not return the source code, but the compiled code. For example if I have a reference to Jackson's ObjectMapper the text content of its PsiTarget/PsiClassImpl is:
> 
> Ahh, ok. In that case, can we include the functionality only if the user has downloaded the source code?

For library code, yes. I am experimenting with triggering the download source action programatically, but I don't think thats a good solution. Perhaps we put a hint in the settings that its benefitial to enable Maven/Gradle auto-download of the sources, then all library sources that are available would be present at the time of the code-completion.
I also think this is just a problem with Java, since there is a separation between the compiled `.jar` and the `sources.jar`. If you e.g. go to the definition of some library code in JavaScript, or GoLang you end up directly in the sources. But I do not have an overview of which languages/package-managers have compiled library code and sources separately.
For now we could simply check if [ClsMethodImpl.sourceMirrorMethod](https://github.com/JetBrains/intellij-community/blob/57f19ccb362b9dedc4fb374c58cf2bfd756a3b3a/java/java-psi-impl/src/com/intellij/psi/impl/compiled/ClsMethodImpl.java#L247)/[ClsClassImpl.getSourceMirrorClass](https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/impl/compiled/ClsClassImpl.java#L507) is present, if yes, we include the library code, if not, we ignore it?

File: src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt (hunk 5 above)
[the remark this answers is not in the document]
Comment (PhilKes) [reply src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt:30]: Okay, I guess we can try out some simple template like:
```
# file_x.py:
<code snippet>

# file_y.py:
<code snippet>

# file_z.py:
<|fim_prefix|> ${infillDetails.prefix} <|fim_suffix|>${infillDetails.suffix} <|fim_middle|>
```

Hopefully at some point there will be a successor of the `gpt-3.5-instruct` model with better support for multi-level completions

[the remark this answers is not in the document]
carlrobertoh: > I think in general the decompiled code should still be usable for the LLMs, although obviously the actual source code would be better.

I think it's fine, let's keep it for now.

> A major concern is how we limit the prompt/context size, especially when Classes are referenced, can we somehow evaluate what parts of the class is relevant? Do we limit the amount of referenced classes/methods?

This is difficult. I think we need to set some maximum token limit boundaries, and even without the class references, some of the context may be left out or even worse, cut out. So, there needs to exist some kind of reranking mechanism to filter out irrelevant data. For now, could we just filter out the class elements?

> Also you can see that since in the FIM prompt itself there is only the method getMoviesSortedByPlaytime, the class field movieService is not present anymore, so the LLM would probably give a faulty/unuseful completion.

Would it be possible to provide the class name or file path from which the `getMoviesSortedByPlaytime` method is coming? For example, instead of passing in the list of `PsiTargets`, you could forward a mapping where the key points to class details and the value to its methods/functions?

carlrobertoh:
> For now we could simply check if [ClsMethodImpl.sourceMirrorMethod](https://github.com/JetBrains/intellij-community/blob/57f19ccb362b9dedc4fb374c58cf2bfd756a3b3a/java/java-psi-impl/src/com/intellij/psi/impl/compiled/ClsMethodImpl.java#L247)/[ClsClassImpl.getSourceMirrorClass](https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/impl/compiled/ClsClassImpl.java#L507) is present, if yes, we include the library code, if not, we ignore it?

Yes, that would work. We can focus on it in a later phase.

> There also IntelliJ plugins providing more specific PSI extensions for frameworks such as Spring, VueJS, Angular,... that we can add later on

Not sure what the impact on the extension's size would be when integrating with all of these plugins.

PR #587, opened 2024-05-31
Draft: feat: Support Codestral 22B 32K

reneleonhardt:
https://mistral.ai/news/codestral/
https://huggingface.co/bartowski/Codestral-22B-v0.1-GGUF
https://huggingface.co/lmstudio-community/Codestral-22B-v0.1-GGUF
https://ollama.com/library/codestral

Still downloading... can you try if prompt and FIM work?
As always the hardest part is figuring out the templates and (stop) tokens, every source says something different...
Would be too easy if there would be standardized model cards containing the info needed to run and use the model, guess we need much more AI to figure that out 🤦

The license is "for research and testing purposes", is it possible to merge?

### DIFF
--- a/src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt
+++ b/src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt
@@ -34,7 +34,13 @@ enum class InfillPromptTemplate(val label: String, val stopTokens: List<String>?
         override fun buildPrompt(prefix: String, suffix: String): String {
             return "<｜fim▁begin｜>$prefix<｜fim▁hole｜>$suffix<｜fim▁end｜>"
         }
-    };
+    },
+    CODESTRAL("Codestral", listOf("</s>")) {
+        override fun buildPrompt(prefix: String, suffix: String): String {
+            return "[SUFFIX]$suffix[PREFIX] $prefix"
+        }
+    },
+    ;
 
     abstract fun buildPrompt(prefix: String, suffix: String): String
 
--- a/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
+++ b/src/main/java/ee/carlrobert/codegpt/completions/HuggingFaceModel.java
@@ -1,5 +1,6 @@
 package ee.carlrobert.codegpt.completions;
 
+import static ee.carlrobert.codegpt.completions.HuggingFaceModel.Model.CST;
 import static ee.carlrobert.codegpt.completions.HuggingFaceModel.Model.P3M;
 import static ee.carlrobert.codegpt.completions.HuggingFaceModel.Model.SC3;
 import static ee.carlrobert.codegpt.completions.llama.LlamaModel.getDownloadedMarker;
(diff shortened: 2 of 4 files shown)


src/main/kotlin/ee/carlrobert/codegpt/codecompletions/InfillPromptTemplate.kt
```diff
@@ -34,7 +34,13 @@ enum class InfillPromptTemplate(val label: String, val stopTokens: List<String>?
         override fun buildPrompt(prefix: String, suffix: String): String {
             return "<｜fim▁begin｜>$prefix<｜fim▁hole｜>$suffix<｜fim▁end｜>"
         }
-    };
+    },
+    CODESTRAL("Codestral", null) {
+        override fun buildPrompt(prefix: String, suffix: String): String {
+            return "[PREFIX]$prefix, [SUFFIX]$suffix"
```
JohanAR, line 40: Based on testing with Mistral's official v3 tokenizer I think the correct template should be: `"[SUFFIX]$suffix[PREFIX] $prefix"`

reneleonhardt: Thank you very much, trying it now!
If you want you could try this branch yourself, would be very helpful 🙂

```bash
git clone https://github.com/reneleonhardt/CodeGPT.git --single-branch --branch support-codestral-22B
java -version # JDK 17+ needed
cd CodeGPT
./gradlew build -x test && ./gradlew prepareSandbox runIde
```
* Switch `Settings | Tools | CodeGPT | Providers` to `LLaMA C/C++ (Local)`
* Open `LLaMA C/C++ (Local)` Provider settings
* Select `Codestral 22B` (3-bit is the smallest download with 10.8 GB)
* Start server
* Open Chat or try inline code-completions (enabled by default)

I added `PromptTemplateTest`s, could you double-check that the template output is as expected/correct for Codestral?
So, is the template really identical to Mixtral? 🤔

Until now the chat seems amazing, no hickups or problems, the answers are very high quality ❤️
Although on my machine (M1 Max 64 GB) the recommended Q6_K (18.3 GB file) is slow, but not unusable like other big models... 😅

reneleonhardt: All "model cards" are different, https://ollama.com/library/codestral/blobs/5dea4f4d0fff lists `[MIDDLE]` for example as possible stop token... sigh

JohanAR: I don't think I can't run codestral locally (I only have 12GB vram) but I do have an API key if the same branch ought to be usable with Mistral's api endpoints.

Mistral-7b and Mixtral-8x7b both use Mistral V1 tokenizer, while Codestral and Mixtral-8x22b use Mistral V3 tokenizer. Chat templates are mostly the same, with some subtle differences :) I did some experiments with them earlier and I think I've got a fairly good idea of how they're supposed to work. I'll join the discord and maybe it's easier to talk about it in real time.

If you wanna check out my experiment results I posted them in this conversation on HF https://huggingface.co/bartowski/Codestral-22B-v0.1-GGUF/discussions/1#66598a7889ed093982767199

JohanAR: > All "model cards" are different, https://ollama.com/library/codestral/blobs/5dea4f4d0fff lists `[MIDDLE]` as possible stop token for example... sigh

I haven't had had much opportunity to test codestral yet, but I think if it's prompted correctly it will probably be enough to use the `</s>` stop token. If the other special tokens show up unexpected I think it's generally a sign that there is some problem with the model (common with merges and some community fine-tunes, especially where they change the original prompt format) or that it's being fed the wrong prompt format.

reneleonhardt: Wow, I never saw how to get the actual templates, thank you for contributing those insights!
Yeah, many models are not working for me even if the templates supposed to be correct, I suspect that it's a healthy mix of wrong models and templates... not easy to develop correctly against all those permutations 😞

I still don't understand why templates are being changed when base models are converted to GGUF, and why this is invisible in the model card??
If I would download a converted Mixtral-8x22b I would expect the templates to be identical.
But I still can't even decipher most base model tokenizer JSONs, they're always different and incomplete (some use special tokens like BOS or EOS, some not?) and even in code completion models I can't find FIM templates... didn't know how you guys do that until now, what kind of "standard" is that? 🤷

I only contributed local llama.cpp models to CodeGPT, API models require much more integration like form settings I suspect, sorry... 😅

JohanAR: It's a pretty new technology, so no settled standards yet, and as soon as someone tries to standardize something they realize that it won't work for the latest stuff so they have to change it :D Regarding templates in GGUFs, it's up to the person doing the quant to include the correct jinja template. In many recent models the intended template is included in jinja format so it should be pretty straightforward to copy it, but Mistral seems to have abandoned jinja in favour of shipping their own tokenizer (which I think some Chinese models also were doing, so they aren't the first to do that).. Well, it's complicated..

I basically installed the mistral_common Python package ran the tokenization examples from the docs in an interactive prompt and tested different things. Check out https://docs.mistral.ai/guides/tokenization/ if you wanna try it.

PR #601, opened 2024-06-29
feat: fast code edits

carlrobertoh:
# Introducing Fast Code Edits

https://github.com/carlrobertoh/CodeGPT/assets/15891408/ff049e54-daa0-4ca8-84ea-6bb4c836a6d6

We're moving into an era where natural language programming is gaining more attention. Therefore, I'm proud to introduce a new feature called **Fast Code Edits**. This feature is intended for users who want to edit their code directly via prompts.

The feature will be available for CodeGPT users first and can be accessed in two different ways:
- Via CodeGPT's context menu
  context-menu

- Via the editor's floating toolbar
  floating-toolbar


michaelmeirsman:
Nice feature :) thank you for your time on this plugin, it definitely rocks!

carlrobertoh:
> Nice feature :) thank you for your time on this plugin, it definitely rocks!

Thank you for your feedback! This feature is still in beta, and I have more improvements planned :)

carlrobertoh:
Sharing a blog post link for others to reference: https://codegpt.ee/blog/introducing-fast-edits

tvdiepen-reconi:
Ah, is this really only for the payed version? Thats such a shame!

tvdiepen-reconi:
If that is the case, at least let us disable the feature so that we don't have an option in our context menu that doesn't work!

carlrobertoh:
@tvdiepen-reconi 

I'm sorry you feel that way. This feature is available through CodeGPT's own provider, and it doesn't require you to pay in order to use it.

As stated on the website, we do give early access to new features to our own customers first, before launching them to others.

Hang tight, we plan to roll this out to other providers soon.

PR #667, opened 2024-08-20
Fix Codestral Infill templates for code completion in llama.cpp

pentatonick:
Hi @carlrobertoh,

Thank you for sharing this awesome project! 

Please accept my humble contributions to fix the Codestral model for code completion, for which I use locally in llama.cpp.

Thank you,
Nick


carlrobertoh:
Thank you! I think we should update the `llama.cpp` submodule as well.

pentatonick:
Thanks for the merge

> I think we should update the llama.cpp submodule as well.

It is not as simple as just updating the git submodule, as the llama.cpp fails to build with the latest commit on master. This will require troubleshooting and I should be able to tackle in the future unless someone else beats me to it.

Issue #553, opened 2024-05-16
Add enable/disable option for providers


AnonTester:
### Describe the need of your request

I'm using the plugin with a local openai compatible provider and do not want any of my code or code parts to be sent to any external provider. It's been the second time now that after an update of the plugin, my selected provider was reset and and I only noticed because I got an error message pop up that code completion isn't allowed without auth token. The latest update changed the selected provider to Ollama (Local) from my previous Custom OpenAI setting.

### Proposed solution

Add toggles in the settings to enable/disable the AI providers that one wants to use. This would reduce the provider selection in the sidebar chat window to only the enabled ones.

Checkboxes on this setting screen to enable each provider would do the job just fine:
![image](https://github.com/carlrobertoh/CodeGPT/assets/40003252/4296c005-58d7-4631-bfb3-419680288ec2)

### Additional context

Related to this issue is the reset of settings or selected provider in an update - which should not happen. If there are changes that prevent retention of the selection, then I suggest to disable ALL providers by default rather than risking sensitive code or questions to be sent to an external provider, potentially causing data protection issues.

Thank you very much for this plugin!

Rikj000:
> ### Additional context
> 
> Related to this issue is the reset of settings or selected provider in an update - which should not happen. If there are changes that prevent retention of the selection, then I suggest to disable ALL providers by default rather than risking sensitive code or questions to be sent to an external provider, potentially causing data protection issues.

I second this,   
it's very annoying that after each update of CodeGPT my `Selected provider`   
(= Custom OpenAI, for a locally hosted model),   
is reset back to an online option.

I do not want an online connection with any LLM,   
since I do not trust them with my data, nor do I want them to train upon my data.

Please resolve this.

ralfstrobel:
I second this request as well. The better visual separation from #681 is good but honestly only made the menu even larger. Especially for users that have only one dedicated local or cloud provider enabled, there is not even a need for the menu to be cascading.

seb-koch:
+1 - we just recognized this limitation which keeps us from proceeding with CodeGPT. We were planning to continue with the Teams subscription and use Azure OpenAI. We can't use CodeGPT in our professional setting without the control of the model that our engineers use. The risk is too high that the wrong model is used and data is leaked.

PR #681, opened 2024-09-07
feat(ui): hide services with missing api key & models not in users pricing plan

GenericMale:
this hides all the services in the ModelCB which require an API key but don't have one configured.
also only the CodeGPT models which are in the users current PricingPlan are shown.

image

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/toolwindow/chat/ui/textarea/ModelComboBoxAction.java
+++ b/src/main/java/ee/carlrobert/codegpt/toolwindow/chat/ui/textarea/ModelComboBoxAction.java
@@ -95,9 +95,10 @@ public class ModelComboBoxAction extends ComboBoxAction {
     if (availableProviders.contains(CODEGPT)) {
       actionGroup.addSeparator("CodeGPT");
       actionGroup.addAll(getCodeGPTModelActions(project, presentation));
+      actionGroup.addSeparator();
     }
+    actionGroup.addSeparator("Cloud Providers");
     if (availableProviders.contains(OPENAI)) {
-      actionGroup.addSeparator("OpenAI");
       var openaiGroup = DefaultActionGroup.createPopupGroup(() -> "OpenAI");
       openaiGroup.getTemplatePresentation().setIcon(Icons.OpenAI);
       List.of(
@@ -109,10 +110,9 @@ public class ModelComboBoxAction extends ComboBoxAction {
       actionGroup.add(openaiGroup);
     }
     if (availableProviders.contains(CUSTOM_OPENAI)) {
-      actionGroup.addSeparator("Custom OpenAI");
       actionGroup.add(createModelAction(
           CUSTOM_OPENAI,
-          ApplicationManager.getApplication().getService(CustomServiceSettings.class)
+          "Custom: " + ApplicationManager.getApplication().getService(CustomServiceSettings.class)
               .getState()
               .getTemplate()
               .getProviderName(),
@@ -120,7 +120,6 @@ public class ModelComboBoxAction extends ComboBoxAction {
           presentation));
     }
     if (availableProviders.contains(ANTHROPIC)) {
-      actionGroup.addSeparator("Anthropic");
       actionGroup.add(createModelAction(
           ANTHROPIC,
           "Anthropic (Claude)",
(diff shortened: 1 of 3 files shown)


carlrobertoh:
Thank you for contributing!

Could we please revert the models that were displayed for the CodeGPT provider, regardless of the user's pricing plan? The reason for this is that I would like to make sure which SOTA models are provided by the CodeGPT API as soon as the model dropdown is launched.

GenericMale:
done. the model list was rather confusing to me, that's why i tried to clean that up 😅

this is which models are shown based on the pricing plan:
- **Always**: gpt-4o, gpt-4o-mini, claude-3.5-sonnet
- **Never**: deepseek-coder-33b, wizardlm-2-8x22b
- **Anonymous**: llama-3-8b
- **Free**: codellama:chat, mixtral-8x22b
- **Individual**: claude-3-opus, dbrx
- **Anonymous** & **Individual**: llama-3.1-405b, deepseek-coder-v2
- **Free** & **Individual**: llama-3-70b

and this doesn't match with the list i get from `https://codegpt-api.carlrobert.ee/v1/users/details`.

maybe you could clean that up a little bit and maybe have something like a Locked & Unlocked section in the ComboBox.
i fully understand that you always wanna show which models users can subscribe to.

oh also i think it might be problematic that the userDetails from your api isn't always loaded when the CB is filled. if a user is on the free or individual plan he might not get to select his preferred model.

carlrobertoh:
Yeah, I need to go through this integration again. The goal was to offer new models as soon as they're launched without needing to upgrade the plugin itself.

I understand that the popup has grown annoyingly large, especially if you have downloaded dozens of Ollama models. However, the issue with this solution is that if you haven't checked the option to remember the keystore key/value mapping (or can't for whatever reason), then each time you boot up your IDE and open the model combobox, you will be prompted to enter your system password up to four times. I'm not sure how big of an issue this is.

Maybe instead of hiding the providers based on whether the API key is set, we could organize them in the menu better to save some space. For example, we could create multiple groups where each group represents a provider with an extensible submenu containing the models. What do you think?

GenericMale:
submenus sound good to be.
what also would help is implementing #553 by adding an enabled checkbox in all the provider setting screens and by default only have CodeGPT enabled (maybe not let users disable CodeGPT for promo reasons if you want 😉).

carlrobertoh:
Yep, I'm aware of that ticket. I think once we lift them into separate groups, then this provider enabling/disabling will no longer be required.

Let's do it like this: leave the CodeGPT models as is and lift the rest into their own groups. The change should be relatively simple:

```java
var openaiGroup = DefaultActionGroup.createPopupGroup(() -> "OpenAI");
openaiGroup.getTemplatePresentation().setIcon(Icons.OpenAI);
List.of(
        OpenAIChatCompletionModel.GPT_4_O,
        OpenAIChatCompletionModel.GPT_4_O_MINI,
        OpenAIChatCompletionModel.GPT_4_VISION_PREVIEW,
        OpenAIChatCompletionModel.GPT_4_0125_128k)
    .forEach(model -> openaiGroup.add(createOpenAIModelAction(model, presentation)));
actionGroup.add(openaiGroup);
```

If this becomes an issue, we can tackle it later.

GenericMale:
image

i would still prefer if we could somehow hide the services which aren't configured but it's an improvement 😅

carlrobertoh:
I added a few minor changes:

combobox-menu

I think separating them by cloud vs local gives a clearer understanding of what is being offered and making it easier for users to navigate.

GenericMale:
oh yeah definitely, great change!

GenericMale:
oh btw one minor thing i tried to figure out but couldn't (because of my total lack of plugin dev experience):
right now the submenus need a click and most jetbrains menus already open on mouse over.
probably an easy thing to change...

carlrobertoh:
Good question. I'm not sure if this is easily configurable via a single parameter. The `com.intellij.openapi.ui.popup.ListPopup` has an option to configure it via `setShowSubmenuOnHover`, but I'm unsure if this list is used by the combo box.

GenericMale:
ah yeah apparently it is [ComboBoxAction.java#L80](https://github.com/JetBrains/intellij-community/blob/cf365294de1809e4cfed5c6d464aa8a8a3a5540d/platform/platform-api/src/com/intellij/openapi/actionSystem/ex/ComboBoxAction.java#L80).

so i guess this should work in `ModelComboBoxAction`:
```java
@Override
protected JBPopup createActionPopup(DefaultActionGroup group, @NotNull DataContext context, @Nullable Runnable disposeCallback) {
  ListPopup popup = (ListPopup) super.createActionPopup(group, context, disposeCallback);
  popup.setShowSubmenuOnHover(true);
  return popup;
}
```

PR #749, opened 2024-11-01
feat: add optional host to Anthropic settings

alexbfr:
I am using Anthropic's API over AWS bedrock, and to access a anthropic api -> bedrock proxy, I  had to add the ability to change the API url. A nicer way would be to  add another sub-setting like Custom OpenAI, but I didn't have enough time to do so. This PR simply adds another optional field `Host` to the Anthropic settings, which may be left empty. Feel free to dismiss this PR without any explanation.

Great extension, by the way! Kudos!

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,7 +4,7 @@ pluginGroup = ee.carlrobert
 pluginName = CodeGPT
 pluginRepositoryUrl = https://github.com/carlrobertoh/CodeGPT
 # SemVer format -> https://semver.org
-pluginVersion = 2.16.2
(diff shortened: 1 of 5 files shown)


carlrobertoh:
Thanks!

Can we remove the version bump? Otherwise it looks good.

borgotta:
Hi! Didn't notice this PR exists and made a similar one: https://github.com/carlrobertoh/ProxyAI/pull/948
The difference is that I hade to change x509trustmanager in default client because in my case I faced an issue where my company's host was protected by internal CA and default trustmanager doesn't load any additional CAs that may be present in your system.

PR #772, opened 2024-11-19
Use the Ollama generate API for FIM templates

boswelja:
The Ollama `/api/generate` API supports prefix + suffix now, let's use that.

Note that there isn't currently a proper API for context files, I opened https://github.com/ollama/ollama/issues/7738 off the back of this.
For now, I have hardcoded the deepseek-coder context format, and it seems to work OK across codeqwen2.5 and deepseek-coder, though it's probably not ideal

Waiting for https://github.com/carlrobertoh/CodeGPT/pull/771


carlrobertoh:
It looks like none of the models I previously downloaded support this new logic (except the models you listed). Could we keep the FIM template configuration and make it optional instead?

boswelja:
Seems reasonable - it's probably a good idea to also open an issue with ollama for those models, since it sounds like they're not retroactively updating model templates

src/main/kotlin/ee/carlrobert/codegpt/codecompletions/CodeCompletionRequestFactory.kt
```diff
@@ -114,14 +114,21 @@ object CodeCompletionRequestFactory {
 
     fun buildOllamaRequest(details: InfillRequest): OllamaCompletionRequest {
         val settings = service<OllamaSettings>().state
+        val prompt = if (details.context == null || details.context.contextElements.isEmpty()) {
```
carlrobertoh, line 117: At the moment, the `details.context` will always be null and it is not configurable from settings either.

PR #897, opened 2025-02-25
Psi structure for chat

aaudin90:
I made the blanks and refactored the tags. I removed tag duplication when the file is added to the context through the add window and simultaneously opened in the editor.

Add psi structure to chat.

With PSI:

https://github.com/user-attachments/assets/eb7a183a-f158-451d-8d65-209eba89ac6c

Without PSI:

https://github.com/user-attachments/assets/6cbf2993-ea72-4105-a8ec-c493335a9838


jojo2357:
I really like the principle, but I am not sure it is working yet. Did some debug and I don't see the psi serialization ending up in the POST request.

can share steps if my help would be welcome \:)

aaudin90:
> I really like the principle, but I am not sure it is working yet. Did some debug and I don't see the psi serialization ending up in the POST request.
> 
> can share steps if my help would be welcome :)

Thank you for your review! This is still a work in progress. We plan to add serialization functionality and also allow users to manage adding tags to chats from the open editor tabs.

I think it will be ready in a week.

jojo2357:
Let me know if you would like some testing and I would be glad to help

aaudin90:
> Let me know if you would like some testing and I would be glad to help

You can try :) 

Now support only ~~CodeGPT~~ ProxyAI providers and CustomOpenAI.

No access to other models.

jojo2357:
> Now support only CodeGPT ProxyAI providers and CustomOpenAI.

I only have access to Ollama. I can confirm that it does not work with ollama which, looking at the code, makes sense. Are there plans to bring it to the openai suite of providers?

aaudin90:
I'll try to do it today.

aaudin90:
> I only have access to Ollama. I can confirm that it does not work with ollama which, looking at the code, makes sense. Are there plans to bring it to the openai suite of providers?

Dode

PR #946, opened 2025-03-26
feat: clear chat tag by keyboard shortcut

Violine:
Clear active chat tag by keyboard shortcut - "Ctrl+Alt+Shift+T".

https://github.com/user-attachments/assets/14319d43-7a15-4930-a702-6efa62ec8f23


rowi123:
i like this idea, but i don't want to remember keys
in the GUI you have a + then the options with 'x'  also add a remove all next to +

PR #948, opened 2025-03-26
feat: custom base host support for anthropic privider

borgotta:
### Why
Certain organizations use their own proxies for AI tools. Multiple tools, including Cline, offer support for custom hosts/URLs to accommodate these proxies.

### What
- add custom host setting
- use jetbrains-provided tls bundle

The default okhttp3 bundle does not include additional certificates that might be installed on your system, so utilizing the JetBrains-provided TLS bundle ensures all necessary certificates are considered.

Screenshot 2025-03-26 at 19 22 58


borgotta:
https://github.com/carlrobertoh/ProxyAI/issues/697

carlrobertoh:
Many thanks!

PR #989, opened 2025-04-13
feat: use GoogleCompletionRequest.systemInstruction for system prompt

PhilKes:
Depends on https://github.com/carlrobertoh/llm-client/pull/61

- Uses Google API's `systemInstruction` field for system prompts


PhilKes:
Since I havent been following the development of this plugin the past months, I am not sure but I am guessing up until now thought tokens of reasoning models are also just printed to the user in the response, right? 
https://github.com/carlrobertoh/llm-client/pull/61 adds the `thought` fields to the response contents as well, so you could possibly mark the thought content response appropriately, like OpenAI or Google AI Studio does in the chat.

carlrobertoh:
IIRC the thinking output is never displayed. At least when I initially introduced the Gemini thinking models, the support wasn't there yet (or was intentionally removed or something).

PhilKes:
> IIRC the thinking output is never displayed. At least when I initially introduced the Gemini thinking models, the support wasn't there yet (or was intentionally removed or something).

Well currently a response from e.g. Gemini 2.5 Pro looks like this:



Sometimes the thinking output seems to be in the same code block as the answer.
I guess it would be best to use the new `thought` boolean to distinguish thought tokens from the actual completion

PhilKes:
Can you release `llm-client` v0.8.39 so we can merge this? 🙂

PhilKes:
Solved the conflicts

Issue #733, opened 2024-10-12
Provide an Option to Choose Separate Models for Code-completion and the Chat when using Ollama provider


alexandertsukanov:
### Describe the need of your request

Currently, when using the Ollama provider, users are limited to a single model for both code-completion and chat functionalities. This feature request proposes the addition of an option to select different models for these two distinct tasks. By allowing users to choose specialized models for each function, we can enhance the performance and accuracy of both code-completion and conversational interactions for local models.

### Proposed solution

 Introduce a settings panel where users can select their preferred models for code-completion and chat independently.
![CleanShot 2024-10-12 at 11 37 50@2x](https://github.com/user-attachments/assets/f9cf0695-6086-404a-b584-a17b12b94095)
It seems that there are some options available in the screenshot above. However, I am unable to switch to a specific model for code completion while using a different one for chat.

I can only to switch both.

jurejj:
This really is something desperately needed. 

Code completion models are typically much different(smaller, faster) from the ones that are used for chat  and absence of this feature essentially forces you tho chose either one or the other.

I personally would even go further and even make it possible to specify different hosts for these two cases:

My Example.
- dev laptops typically lack powerfull gpus but can still quite happily handle small fast models, needed for code completion while it is struggling with the big ones for chat.
- I personally have an additional gaming machine with appropriate gpu that sleeps during the workday. I would really lahk to use that one to run as big models as possible for chat.

Such setup also avoids the load model time delays when switching model. That would be gold for me.

Issue #1025, opened 2025-05-16
Use different models for chat and code completitions


Prikalel:
### Describe the need of your request

Is it possible to use for example opan ai compatable server for chat 
but for auto code completitions use local llm ?

keyute:
if your local llm supports openai completion endpoint you can just call that directly under code completions

Issue #804, opened 2024-12-18
Separate Settings for Chat Providers and Code Completion Providers


moritzfl:
### Describe the need of your request

Currently, you can only mix and match providers by using the "Custom OpenAI" Provider. 

It would be pretty awesome to have code completion separate from your normal chat service.

Possible scenarios:
- Run one of the models remotely and one locally
- Use Azure for chat but rely on Mistral for code completions
- Mix and match between other providers

If we want to improve usability for the configuration, I'd also make providers appear as a list where you can add or remove providers both for code and chat.

I hope this illustration transports the idea:
![grafik](https://github.com/user-attachments/assets/6a32aa7e-2d26-49e8-828c-a71883a0fac9)

If you think this is a good  direction for this project and the UI, I would be willing to work on it.

PeterCassetta:
To add onto this request, it would be great to be able to select different models within providers for code completion vs chat. (use case: fast code completion, but more robust chat)

abatkin:
I think this is a must-have. I would imagine that everyone always wants a different model for completions vs chat.

PR #1032, opened 2025-05-22
Add separate model selection for code completion

muravvv:
Add separate settings for code completion model: 

- Separate selection of model provider for code completion to use different providers for chat and code completion (for example to use ProxyAI model for chat and other actions + local code completion)
- Separate Ollama model selection for code completion

This pull request closes issues #733, #804 and #1025.

Also during addition of the test for new Ollama model selection, I have found a bug in code completion tests: code completion cache (`CodeCompletionCacheService`) was not resetted between tests. As I used different model responce for same input (`rivate void main` instead of `ublic void main` in other tests), next executed tests (`test code completion with OpenAI provider` and `test code completion with ProxyAI provider`) became failed. To fix this bug I have added cleaning of `CodeCompletionCacheService` to each test.

This bug most likely led that some of the other tests did not work: actually that tests was tested code completion cache instead of real model query code. So after my fix this tests will start working as expected.

### DIFF
--- a/src/main/java/ee/carlrobert/codegpt/settings/GeneralSettings.java
+++ b/src/main/java/ee/carlrobert/codegpt/settings/GeneralSettings.java
@@ -8,6 +8,7 @@ import ee.carlrobert.codegpt.CodeGPTKeys;
 import ee.carlrobert.codegpt.completions.HuggingFaceModel;
 import ee.carlrobert.codegpt.completions.llama.LlamaModel;
 import ee.carlrobert.codegpt.conversations.Conversation;
+import ee.carlrobert.codegpt.settings.service.ModelRole;
 import ee.carlrobert.codegpt.settings.service.ProviderChangeNotifier;
 import ee.carlrobert.codegpt.settings.service.ServiceType;
 import ee.carlrobert.codegpt.settings.service.anthropic.AnthropicSettings;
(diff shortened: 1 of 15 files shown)


muravvv:
Yes, I also don't like resulting UI, but I have no idea how to make it better. 

My first idea was to duplicate whole `Providers` category: create `Code completion providers` with its own `Selected provider` setting and pages for each provider:
separate category
But significant number of parameters (API keys, hosts and so on) actually are common for all actions with separate models. And it is not good to requere a user to change same API key in five places, when he's API key for one of providers was changed.

A good settings page is used in Continue plugin:
continue plugin
In that variant model for each task is selected from unified list for all models from all providers (in their plugin this list is defined manually in yaml config file). But it cannot be used in ProxyAI as for some providers model name is entered as text, and so complete list of available models cannot be built.

So if there is an idea how to make a good UI, I can implement it (if I'll have time for this, of course). Expectly as new Auto Apply is a great feature and it also requeres separate model (as many good models cannot generate valid diff blocks).

carlrobertoh:
Yeah... I actually like Continue's approach and I think we could implement something similar. We could group all the models together by their respective providers and display them for each available feature as follows:
              
- ProxyAI · GPT-4.1  
- ProxyAI · Claude Sonnet 4  
...  
- OpenAI · GPT-4.1  
...  
- Anthropic · Claude Sonnet 4  
...  
- Ollama · qwen3-14b

I'm going to merge this as is, but I'll likely put my hands on it before the release.

carlrobertoh:
Perhaps something like this:
models-settings

Issue #985, opened 2025-04-10
Custom OpenAI settings are overwritten by the defaults


klocus:
### What happened?

It seems to me that the problem only occurs when we add more than one custom OpenAI APIs.

I uninstalled the plugin, cleaned up the files after it. There is no longer this deprecated file now, and still, with more than one custom Open AI this bug occurs.

WebStorm 2024.3.5

### Relevant log output or stack trace

```shell

```

### CodeGPT version

3.2.2-241.1

### Operating System

macOS

PR #1051, opened 2025-06-13
Fix Custom OpenAI settings

russ-p:
Fixes unintended settings overwrite when setting `templateComboBox` programmatically (issue #985)


carlrobertoh:
This should be fixed. Please reopen, if not.

PR #1090, opened 2025-08-01
fix: Custom Open AI services do not adhere to model selections

### LINKED ISSUE #1089: Multiple Custom OpenAI Providers only uses first definition after Models per Feature update
### What happened?

After the new "Model per Feature" addition, the Custom OpenAI integration stopped working properly.

**What happens:**
When you select a Custom OpenAI provider, the plugin will _always_ use the first defined definition in the Custom OpenAI settings.
This applies to both Model per Feature as well as the model selection below the chat-box.

**What I found:**
I spelunked the ProxyAI sources and found the possible culprit.
1. When you save the custom provider settings, it sets the first definition as `active`.
2. When an AI operation is invoked, after some factories, the current `active` definition is used.

However, the `active` var is *only* ever written to in:
* The settings _(I think to ensure a non-null value)_
* A settings migration _(due to the Models per Feature update)_

So uhm... Yeah, ProxyAI only ever uses the first Custom OpenAI definition.

### Relevant log output or stack trace

```shell

```

### Steps to reproduce

I use LM Studio locally and it supports JIT loading of models via the `model` body parameter.

Prep:
1. Define two Custom OpenAI definitions, both pointing to LM Studio (or something where you can check which model is being used.)
2. Set the providers to use different models.

Repro:
- Switch between models in the combobox underneath the chatbox, in the main UI, during chat. You can see that only ever a single model is used.
- Similarly, switching the model in the Model per Feature, I used the CommitMessage feature (but the abstractions make this bug apply to all features). Here also, switching between definitions within the custom providers has no effect

### CodeGPT version

3.5.0-241.1

### Operating System

Linux

Juraji:
Fixes #1089 

My (second) hunch was correct. The `CustomOpenAIRequestFactory` as well as several points in the Code Completion logic referenced `ee.carlrobert.codegpt.settings.service.custom.CustomServicesState#getActive` but that property always contains the same (first) custom service definition state.

### DIFF
--- a/src/main/kotlin/ee/carlrobert/codegpt/actions/CodeCompletionFeatureToggleActions.kt
+++ b/src/main/kotlin/ee/carlrobert/codegpt/actions/CodeCompletionFeatureToggleActions.kt
@@ -40,8 +40,9 @@ abstract class CodeCompletionFeatureToggleActions(
             }
 
             CUSTOM_OPENAI -> {
-                service<CustomServicesSettings>().state.active.codeCompletionSettings.codeCompletionsEnabled =
-                    enableFeatureAction
+                service<CustomServicesSettings>()
+                    .customServiceStateForFeatureType(FeatureType.CODE_COMPLETION)
+                    .codeCompletionSettings.codeCompletionsEnabled = enableFeatureAction
             }
 
             ANTHROPIC,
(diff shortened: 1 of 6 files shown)


Juraji:
@carlrobertoh I could completely eliminate the `active` property, if you want. Just say the magic word of confirmation! :D

Anyway, let me know if you need any changes or don't like the fix. I'll see what I can do.

PR #1103, opened 2025-08-18
feat: chat-based FIM code completion for Custom OpenAI providers

gmontamat:
@carlrobertoh: this is the only feature missing in ProxyAI to finally ditch Github Copilot!

FIM Code completion is done in ProxyAI using gpt-3.5-turbo-instruct via the `/v1/completions` endpoint only which is now [legacy](https://platform.openai.com/docs/guides/completions). Many OpenAI-compatible APIs do not support this endpoint.

I added Chat-based FIM code completion and tested with gpt-4.1.

May conflict with: https://github.com/carlrobertoh/ProxyAI/pull/1090, I could rebase if necessary.

Screenshot_20250818_130647

Sample GPT-4.1 chat completion:

Screenshot_20250821_114946


carlrobertoh:
Looks good. However, I believe the UI should be made more clearer. I had to follow the exact configuration from your screenshots to quickly understand how to test this logic. I think this approach should be applied to every provider and model that isn't trained on FIM tokens.

PR #1191, opened 2026-02-27
custom openai responses API support

AlexanderLuck:
OpenAI supports new models, especially codex, only via new responses API.
Added support for responses API, tried to stay close to plugins patterns.

test by making new custom gpt configuration for example body:

{
  "stream" : true,
  "max_output_tokens" : 8192,
  "model" : "gpt-5.3-codex",
  "messages" : "$OPENAI_MESSAGES",
  "temperature" : 1
}

using the
https://api.openai.com/v1/responses
url


sapphirepro:
> One question tho: if the plugin's native OpenAI provider supports the Responses API, why would you need to configure it via Custom OpenAI? One reason could be faster access to newer models, but other than that I don't see a very compelling reason, since the Responses API isn't widely supported by 3rd party inference providers.

Even I am not an author of that commit, this question is sensible in fact. Responses api is better for agent. And native OpenAI is insufficient because aside of  older models, it's also impossible to override anything like reasoning effort etc, it's complete black box without an option to chose anything. At minimum: reasoning effort, verbosity, service tier, if to allow native OpenAI web search etc. Having in settings just option to enter API key and hope it works is no go, that's why I have all Open AI models over custom open ai, manually cloning them into various reasoning efforts, service tiers and other things. Also repsonses api is much more token caching efficient

sapphirepro:
> Good news is, since koog is doing the heavy lifting, a bunch of things can now simply be cut, the main commit is now <250 lines for the responses API.

@AlexanderLuck 
I have sent you pull request. Models can now be nicely sorted (screen below). Since I used your version for customization, request is sent to you. Maybe can merge both things at once then and avoid version conflicts
изображение

PR #1195, opened 2026-03-14
Responses API, fixed and added diagnostics, Custom OpenAI model sort (up down buttons)

sapphirepro:
1. Used as base partly broken Agent side Customer OpenAI (Responses), but fixed it. Now there models work, tools call also operational. 
2. Fixed Diagnostics (in chat => "+" => Diagnostics, which previously added random 1 file). Now provides all attached context diagnostics with selection of "errors" only or "all".
3. Added diagnostics tool to agent mode
4. Added sort of Custom OpenAI models, so they can be now manually organized with up/down keys.

Ordinary OpenAI (predefined) remained untouched. Fixed both Custom OpenAI and responses version from broken state in agent mode to operational state.

Please review and if possible accept fully/partly/or modify on need.

Review recommended. For me all works. With kind regards.


carlrobertoh:
Merged and simplified couple of changes.

PR #1224, opened 2026-05-17
Context editor improvements (file/folder picker). Target IDE shift

sapphirepro:
- Target IDE build shifted to 262.* max supported (current EAP).
- Added context settings for file/folder picker to be more usable: File filter settings, max files, max folders, display right options for better views, sorting.
Screenshot_20260517_031530
Screenshot_20260517_031608

Hello. I add screenshots for fast view. I tried my best to for options to keep as it was for you with options to configure for better usability. On screenshot is my best views, where all perfectly sorted, cut fiename from right to see fast what folder, what file etc. Default is your way it was. Simply that way with configuration it's much more convenient. For me 25 files and only open/recent is not enough and without sorting it's nightmare.

Also I shifted max supported build to 262.X for EAP (that I currently use and tested on).

With kind regards. 

PS. If you don't like any text, view, edit as you wish. Edits are allowed by pull request, so you can edit if needed.


sapphirepro:
@carlrobertoh I am sorry for tagging you, but just for request not to get lost

PR #1228, opened 2026-06-10
feat: add action to send selected Problems Tool Window entry to Chat …

bysunlit:
…context

Registers ProxyAI.AddProblemToContext with no default shortcut so users can bind it via Settings -> Keymap. Introduces SingleDiagnosticTagDetails plus its processor, exposes ChatToolWindowTabPanel#addContextTag, and updates PsiStructureRepository's exhaustive when branches.


bysunlits:
Drafted with significant AI assistance. I reviewed the design, verified
the build, and exercised the happy path in a running IDE (selecting an
error in the Problems tool window → shortcut → tag appears in Chat with
the expected diagnostic in the prompt). The two fallback paths in
`extractProblemContext` (OpenFileDescriptor / VIRTUAL_FILE only) haven't
been exercised end-to-end. Happy to revisit the reflection-vs-direct-import
choice or trim the fallbacks if preferred.

